PR #2, opened 2017-12-18
Wdfn 4 - Create base Flask application for waterdataui

mbucknell:
Follow pattern suggested in http://exploreflask.com/en/latest/organizing.html#package and what has been done on previous CIDA/OWI Flask applications


run.py
```diff
@@ -0,0 +1,16 @@
+import argparse
```
danielnaab: Could we add a "#!/usr/bin/env python" shebang line here and "chmod +x run.py"?

mbucknell: I'm reluctant to do that because typically I use the python that has been created in my virtualenv to run this file. On my machine, this would actually use python 2.7.13. @ayan-usgs , your thoughts?

danielnaab: Yeah... I used virtualenvwrapper to activate my environments, but you could run `source env/bin/activate` also - it will make the venv's python the default on your path.

mbucknell: Being the control freak that I am, I don't usually use env/bin/activate. That way I have complete control over which version I'm running.

mbucknell: I can go ahead and make that change since I can still run it my prefered way.

ayan-usgs: While I haven't found shebangs to be particularly useful with respect to how I work, I don't have any objections to including it in the file.

Recognizing that there is a difference they haven't been useful there is a difference between the `#!/usr/bin/env python` as @danielnaab is proposing and `#!/usr/bin/python`, where the former does respect a virtualenv if activated and the latter does not.

The only hesitance I have about adding it is that many *nix systems currently come with both Python 2 and Python 3 and they appear on the path as `python` and `python3` (may or not include minor version) respectively. Outside of a virtualenv, this could cause a script written for Python 3 to be run under Python 2.

danielnaab: Thanks! I see your point on control and knowing exactly what you're running against. What I especially like about activating it is the postactivate script... with virtualenvwrapper I type "workon waterdata" to activate, and the below script is run to do other environment configuration:

```bash
#!/usr/local/bin/zsh
# This hook is sourced after this virtualenv is activated.
cd ~/src/waterdataui
nvm use v8.9.3
```

danielnaab: @ayan-usgs I suppose we could do "#!/usr/bin/env python3.6" to clarify that, and it would still work with venvs.

ayan-usgs: @danielnaab -- Using `#!/usr/bin/env python3.6` sounds good to me.

mbucknell: 👍 . Going to merge

PR #3, opened 2017-12-19
instance config.py may not be present if nothing needs to be overridden

ayan-usgs:


waterdata/__init__.py
```diff
@@ -7,6 +7,9 @@
 
 # Loads configuration information from config.py and instance/config.py
 app.config.from_object('config')
-app.config.from_pyfile('config.py')
+try:
```
ayan-usgs: Wondering if this is an appropriate thing to do... :thinking:

mbucknell: I think it's fine. In the past, we always had tier specific settings, but perhaps we won't, particularly if we are using publically available endpoints

danielnaab: I added a couple config settings for static assets - but you both will probably have feedback on how I set it up. I'm fine with this change though.

ayan-usgs: @mbucknell -- Agreed. If we decide we need specific settings in the future, this doesn't preclude the use of a tier specific config.

jkreft-usgs: and there will be tier-specific config, even if it is just boring stuff like a dev-specific google analytics code

PR #4, opened 2017-12-19
WDFN-7 - Add Javascript and CSS build and development workflow.

danielnaab:
Add Javascript and CSS build and development workflow. Includes's 18F's web design guidelines.

- This adds a new build step for static assets - "npm run build"
- And a watch task - "npm run watch"

Both compile into the /dist directory. "watch" will load a dev server on port 9000, and also start the Flask dev server on 5050 (using the default Python.. so the virtualenv would need to be activated. not sure how to get around that without making that step optional).

I looked into hosting the static assets via Flask (and putting them in waterdata/static) but decided against it for a few reasons:
1) integrating livereload with node.js is easier than Python
2) For day-to-day coding, we'll want a node-based "watch" task - so we'll have two processes running irregardless of which one serves the files up
3) in production, the web server will handle static hosting, so it seems nice to totally keep that concern out of the Python code

This commit also includes some usage of the 18F CSS classes in `base.html`, to demo that they work and all assets (fonts, JS, css, images) are being built properly. Here are current file sizes:

```bash
 dnaab@IGSARMEWLTDJN-M  ~/src/waterdataui   wdfn-7-nodejs-build  ll dist
total 184
  0 drwxr-xr-x   6 dnaab  staff    204 Dec 19 15:56 ./
  0 drwxr-xr-x  23 dnaab  staff    782 Dec 19 15:56 ../
 48 -rw-r--r--   1 dnaab  staff  22376 Dec 19 15:56 bundle.js
  0 drwxr-xr-x  34 dnaab  staff   1156 Dec 19 15:56 fonts/
  0 drwxr-xr-x  59 dnaab  staff   2006 Dec 19 15:56 img/
136 -rw-r--r--   1 dnaab  staff  66897 Dec 19 15:56 main.css
```


package.json
```diff
@@ -0,0 +1,61 @@
...
+  "description": "Compiled static assets for Water Data For The Nation",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "clean": "rm -rf dist && mkdir dist",
+    "build": "npm run clean && run-p build:*",
+    "build:css": "node-sass --include-path node_modules/uswds/src/stylesheets assets/styles/main.scss | uglifycss > dist/main.css",
+    "build:fonts": "mkdir -p dist/fonts && cp node_modules/uswds/src/fonts/* dist/fonts",
+    "build:images": "mkdir -p dist/img && cp -r node_modules/uswds/src/img/* dist/img",
+    "build:js": "browserify -t uglifyify assets/scripts/index.js | uglifyjs -c > dist/bundle.js",
```
danielnaab: `run-p` on the "build" task runs each of the build:* tasks in parallel.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "serve:static": "live-server --port=9000 --cors --no-browser dist",
+    "watch": "run-p watch:* serve:static",
+    "watch:css": "nodemon -w assets/styles -e scss -x \"npm run build:css\"",
+    "watch:fonts": "nodemon -w node_modules/uswds/src/fonts -x \"npm run build:fonts\"",
+    "watch:images": "npm run build:images",
+    "watch:js": "watchify assets/scripts/index.js -o dist/bundle.js -dv",
```
danielnaab: `watchify` uses the same settings as `browserify` (specified below in package.json), but doesn't include the uglify options that are specified (above) on the command line.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "rollupify": "^0.4.0",
+    "uglify-js": "^3.2.2",
+    "uglifycss": "0.0.27",
+    "uglifyify": "^4.0.5",
+    "uswds": "^1.4.4",
+    "watchify": "^3.9.0"
+  },
+  "browserify": {
+    "transform": [
+      "rollupify",
```
danielnaab: If we use ES6 imports, `rollupify` can "tree shake" unused code out of the JS bundle. Unfortunately, the CommonJS modules that the 18F code uses can't benefit from this.

package.json
```diff
@@ -0,0 +1,61 @@
...
+      [
+        "babelify",
+        {
+          "presets": [
+            "env"
+          ]
+        }
+      ]
```
danielnaab: The env preset for Babel will keep us up to date on ECMAScript additions.

waterdata/templates/base.html
```diff
@@ -0,0 +1,50 @@
...
+                    var socket = new WebSocket('{{ config.LIVE_RELOAD_PATH }}');
+                    socket.onmessage = function(msg) {
+                        if (msg.data == 'reload') window.location.reload();
+                        else if (msg.data == 'refreshcss') refreshCSS();
+                    };
+                    console.log('Live reload enabled.');
+                })();
+            }
+            // ]]>
+        </script>
```
danielnaab: This snippet maintains the open connection to the `live-server` instance, that can auto-reload when static assets are recompiled.

waterdata/templates/partials/footer.html
```diff
@@ -0,0 +1,10 @@
+<footer class="usa-footer usa-footer-slim" role="contentinfo">
+    <div class="usa-footer-secondary_section">
+        <div class="usa-grid">
+            <div class="usa-footer-logo">
+                <img class="usa-footer-slim-logo-img" src="{{ config.STATIC_ROOT }}/img/logo-img.png" alt="Logo image">
+                <h3 class="usa-footer-slim-logo-heading">U.S. Geological Survey</h3>
+            </div>
+        </div>
+    </div>
+</footer>
```
danielnaab: There are actually some templates bundled with USWDS that we could possible import instead of making our own.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "build:fonts": "mkdir -p dist/fonts && cp node_modules/uswds/src/fonts/* dist/fonts",
+    "build:images": "mkdir -p dist/img && cp -r node_modules/uswds/src/img/* dist/img",
+    "build:js": "browserify -t uglifyify assets/scripts/index.js | uglifyjs -c > dist/bundle.js",
+    "serve:static": "live-server --port=9000 --cors --no-browser dist",
+    "watch": "run-p watch:* serve:static",
+    "watch:css": "nodemon -w assets/styles -e scss -x \"npm run build:css\"",
+    "watch:fonts": "nodemon -w node_modules/uswds/src/fonts -x \"npm run build:fonts\"",
+    "watch:images": "npm run build:images",
+    "watch:js": "watchify assets/scripts/index.js -o dist/bundle.js -dv",
+    "watch:python": "./run.py"
```
ayan-usgs: I noticed that everything is set to run on port 9000. Perhaps this is ignorance on my part, but should `app.run(host=host, port=5050, threaded=True)` in run.py be changed to `app.run(host=host, port=9000, threaded=True)`, or does it not matter?

README.md
```diff
@@ -20,4 +37,32 @@ any values in the project's config.py.
...
+Node.js dependencies are installed via:
+
+```bash
+npm install
+```
+
+And a development server hosting both the Python and static assets run with:
+
+```bash
+npm run watch
```
mbucknell: It is a little confusing to describe how to run the development server above (line 37) and then to have this. I would like to see the running as a separate section. So recommend removing this and have a separate section for running for local development. which will include activating the virtualenv.

README.md
```diff
@@ -20,4 +37,32 @@ any values in the project's config.py.
...
+```
+
+To build the complete distributable package, built to `./dist`:
+
+```bash
+npm run build
```
mbucknell: Is this a step that would typically be done during the build process on Jenkins? Can you explain how to run locally with built assets? This can be helpful to make sure the entire build process works and to debug issues that only occur with built assets.

package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "waterdataui-assets",
+  "version": "1.0.0",
```
mbucknell: The version should probably match what is in .bumpversion.cfg, currently 0.1.0dev0. Can we use bumpversion to update this version, too?

README.md
```diff
@@ -20,4 +37,32 @@ any values in the project's config.py.
 4. To run the development server, execute the following:
 ```bash
 % env/bin/python run.py
-```
\ No newline at end of file
+```
+
+## Static assets development
+
+Javascript and SASS assets are built against Node.js v8.9.3. Usage of
```
mbucknell: Can we add the install step to dev_install.sh?

package.json
```diff
@@ -0,0 +1,61 @@
...
+  "description": "Compiled static assets for Water Data For The Nation",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "clean": "rm -rf dist && mkdir dist",
+    "build": "npm run clean && run-p build:*",
+    "build:css": "node-sass --include-path node_modules/uswds/src/stylesheets assets/styles/main.scss | uglifycss > dist/main.css",
```
mbucknell: Does this imply that tools like node-sass have been installed globally? That will not be the case particularly when using Jenkins.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "url": "git+https://github.com/usgs/waterdataui.git"
+  },
+  "author": "USGS",
+  "license": "CC0-1.0",
+  "bugs": {
+    "url": "https://github.com/usgs/waterdataui/issues"
+  },
+  "homepage": "https://github.com/usgs/waterdataui#readme",
+  "dependencies": {
+    "babel-core": "^6.26.0",
```
mbucknell: I'm still unclear how package-lock.json works with package.json. Typically we have pinned version numbers of dependencies.

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "type": "git",
+    "url": "git+https://github.com/usgs/waterdataui.git"
+  },
+  "author": "USGS",
+  "license": "CC0-1.0",
+  "bugs": {
+    "url": "https://github.com/usgs/waterdataui/issues"
+  },
+  "homepage": "https://github.com/usgs/waterdataui#readme",
+  "dependencies": {
```
mbucknell: These should all be devDependencies?

package.json
danielnaab: It can stay at localhost:5050 - the Flask hosted app load static files from localhost:9000

package.json
```diff
@@ -0,0 +1,61 @@
...
+    "browserify": "^14.5.0",
+    "live-server": "^1.2.0",
+    "node-sass": "^4.7.2",
+    "nodemon": "^1.13.3",
+    "npm-run-all": "^4.1.2",
+    "rollupify": "^0.4.0",
+    "uglify-js": "^3.2.2",
+    "uglifycss": "0.0.27",
+    "uglifyify": "^4.0.5",
+    "uswds": "^1.4.4",
```
mbucknell: I think this is the only non dev dependency.

README.md
danielnaab: Yes, I was considering that but was concerned that there's more than one way to install node.js and I wasn't sure if it would be more confusing than helpful to put in there. Would we be using dev_install.sh for build and test automation? Or just as a developer helper?

README.md
danielnaab: Good point, will do.

README.md
danielnaab: Yes, that would be the production build. If we wanted to host those locally, there would be an extra step to run the server. I'll add docs for that.

package.json
danielnaab: Hadn't considered that - would be good to do. The dev0 suffix might not be valid with the semantic versioning npm expects - although, we won't actually be publishing to the npm repository. Will investigate.

package.json
danielnaab: It's the locally installed version. I believe the way it works with run scripts is that node adds the bin directory of each package in node_modules/ to the path.

README.md
mbucknell: It's just a developer helper

package.json
mbucknell: In the past that wasn't true but a lot has changed with node since I tried that. As long as it works, I'm happy.

README.md
```diff
@@ -6,6 +6,23 @@ has been developed using Python 3.6. This is a work in progress.
...
+it should contain the following:
+
+```python
+DEBUG = True
+
+# The root location of static assets (Javascript, CSS, Images)
+STATIC_ROOT = 'http://localhost:9000'
+
+# If this is not set, live reload will be disabled.
+LIVE_RELOAD_PATH = 'ws://localhost:9000/ws'
```
mbucknell: How is this used? I'm not finding it in standard Flask and I don't see any new python dependencies.

package.json
jkreft-usgs: possible future pull request on the standards repo?

PR #6, opened 2017-12-27
Wdfn 8 - Add test runners

danielnaab:
This adds Jasmine and Karma, and maps a few new npm run commands:

```
npm run test  # alias for test:karma
npm run test:jasmine
npm run test:karma
npm run test:watch
```


karma.conf.js
```diff
@@ -0,0 +1,69 @@
...
+    // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
+    preprocessors: {
+        'assets/**/*.js': ['browserify']
+    },
+
+
+    // test results reporter to use
+    // possible values: 'dots', 'progress'
+    // available reporters: https://npmjs.org/browse/keyword/karma-reporter
+    reporters: ['progress'],
```
mbucknell: I think you will need some sort of coverage reporter from Coveralls.

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Compiled static assets for Water Data For The Nation",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "test": "npm run test:karma",
+    "test:jasmine": "jasmine assets/tests/**",
```
mbucknell: I'm not familiar with running tests using just a jasmine command. What is the difference between this on using karma?

package.json
```diff
@@ -4,7 +4,10 @@
...
+    "test:karma": "karma start",
+    "test:watch": "karma start --no-single-run --auto-watch",
```
mbucknell: Like this !

package.json
```diff
@@ -38,6 +41,11 @@
     "babel-preset-env": "^1.6.1",
     "babelify": "^8.0.0",
     "browserify": "^14.5.0",
+    "jasmine": "^2.8.0",
```
mbucknell: Should we pin the versions of the npm libraries that we are using?

karma.conf.js
danielnaab: Yes you're right - I'll add that as part of WDFN-9

package.json
danielnaab: That will run via node.js rather than in a browser. I suppose it remains to be seen what our tests will look like and whether that will be useful, but I thought I'd just put it in there for now.

ayan-usgs: That's pretty cool. :thinking:

package.json
danielnaab: They're pinned in package-lock.json - so this here is mostly just useful as a rule with something like npm-check-updates.

package.json
mbucknell: Since our javascript code will always run in a browser we should be using the test:karma. I lean towards removing it unless we have javascript that is not run in a browser to test.

danielnaab: Ok, I'll remove it.

PR #11, opened 2017-12-28
add codacy badge

ayan-usgs:


ayan-usgs:
> It would also be good to explore setting the linters for our preferred style on Codacy. It will use standard configuration files in the repo if available.

Yep. I can take a look at that.

ayan-usgs:
@mbucknell -- I've added configuration files for eslint and pylint.

mbucknell:
Are these just standard settings for js and python?

ayan-usgs:
> Are these just standard settings for js and python?

Mostly. The python config definitely is. The javascript is all of the eslint recommended stuff and one warning that I decided to include regarding brace style.

PR #12, opened 2017-12-28
WDFN-8 - Use Firefox instead of Chrome with Karma

danielnaab:


package.json
```diff
@@ -42,8 +42,8 @@
     "jasmine": "^2.8.0",
     "karma": "^2.0.0",
     "karma-browserify": "^5.1.2",
-    "karma-chrome-launcher": "^2.2.0",
```
mbucknell: I think it's okay to leave karma-chrome-launcher in package.json. When we develop locally it may be more convenient to use Chrome (I know I like their developer tools better). We could even include a version of npm test which runs chrome (maybe test:chrome)

PR #13, opened 2017-12-28
eslint additions

danielnaab:
- Added env settings for CommonJS/node and ECMAScript 6.
- Added "recommended settings" extends
- Enforce consistency with semi-colons and quotation marks


.eslintrc
```diff
@@ -31,6 +37,8 @@
     "no-undef": 2,
     "no-use-before-define": 2,
     "brace-style": [1,"1tbs",{}],
-    "no-mixed-spaces-and-tabs": 2
+    "no-mixed-spaces-and-tabs": 2,
+    "quotes": ["error", "single"],
```
mbucknell: We may have a style fight here. I prefer single quotes. Never mind, I was interpreting this incorrectly. I like single spaces 😄

.eslintrc
```diff
@@ -1,6 +1,12 @@
 {
+  "env": {
+    "commonjs": true,
+    "es6": true
+  },
+  "extends": "eslint:recommended",
```
ayan-usgs: So, I'm wondering if this would allow us to remove a some but not all of the stuff from lines 3 through 33.

danielnaab: Yes - I'll go through them

ayan-usgs: I could also go through them if you would prefer.

danielnaab: No problem, already on it :)

PR #14, opened 2017-12-29
Clear linting warnings from our .pylintrc.

danielnaab:


waterdata/tests/test_views.py
```diff
@@ -1,15 +1,21 @@
...
+"""
 
 from unittest import TestCase
 
 from .. import app
 from ..views import __version__
 
+
 class TestHomeView(TestCase):
+    """Basic tests validating existence of pages."""
```
mbucknell: I'm not sure how necessary doc string comments are in tests. Test case classes should be tied to a particularly module or class and the methods themselves should be small and the function name should be descriptive

danielnaab: I don't disagree, but we could strive to have zero linting errors... it appears to not actually be possible (yet) to have different .pylintrc files in subdirectories: https://github.com/PyCQA/pylint/issues/618

mbucknell: We could try adding no-docstring-rgs: '^test\w|^Test\w' (I think that would work. Not the best at regex)

danielnaab: Ah! I didn't know we could do that in pylint. I'll push another commit.

PR #15, opened 2018-01-02
Monitoring location view

ayan-usgs:
I implemented (at least I think I implemented) the error handling that was discussed in our slack thread.


requirements.txt
```diff
@@ -1,6 +1,17 @@
...
 click==6.7
 Flask==0.12.2
+idna==2.6
 itsdangerous==0.24
 Jinja2==2.10
 MarkupSafe==1.0
-Werkzeug==0.13
\ No newline at end of file
+numpy==1.13.3
+pandas==0.21.1
```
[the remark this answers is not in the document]
mbucknell: Can we use the json format rather than rdb? Gut reaction is that pandas is overkill but I haven't looked at the code yet.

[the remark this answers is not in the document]
ayan-usgs: > Can we use the json format rather than rdb?

I could be misreading, but I don't think water services supports json: https://waterservices.usgs.gov/rest/Site-Service.html#format.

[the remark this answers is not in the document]
mbucknell: Shoot you are right. I still think pandas is overkill. I

waterdata/templates/monitoring_location.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'base.html' %}
+
+{% block content %}
+    {% if status_code == 200 %}
+        <h3>Station Name: {{ station_name }}</h3>
```
mbucknell: Semantically we should probably be using h1.  @jkreft-usgs, opinion?

waterdata/views.py
```diff
@@ -5,9 +5,40 @@
...
+        return render_template('monitoring_location.html',
+                               status_code=status,
+                               station_name=station_name
+                               )
+    elif 400 <= status < 500:
+        return render_template('monitoring_location.html',
+                               status_code=status,
+                               reason=reason
+                               )
+    elif status == 500:
```
mbucknell: Do we know what we can potentially get from the site service? This would imply that only a 500 would have us set a 502?

waterdata/tests/test_utils.py
```diff
@@ -0,0 +1,100 @@
...
+        with mock.patch('waterdata.utils.r.get') as r_mock:
+            r_mock.side_effect = r.exceptions.Timeout
+            result = get_water_services_data(self.test_hostname, 'nwis/site/?site={}'.format(self.test_site_number))
+        expected = (None, None, None)
+        self.assertTupleEqual(expected, result)
+
+
+class TestParseRdb(TestCase):
+
+    def setUp(self):
```
mbucknell: Should also test an empty string and one that doesn't provide any data lines. Should also have a test that has more than one data row (we will get rdb's like this for site's which are only unique by agency code). For really completeness might want to include a count mismatch between the column header row and a data row.

waterdata/utils.py
```diff
@@ -0,0 +1,49 @@
...
+import requests as r
+
+
+def get_water_services_data(hostname, url_path):
+    """
+    Make a request to USGS water services.
+
+    :param str hostname: Scheme and hostname of the target service
+    :param str url_path: Path
+    :return: Water services data (in RDB format), request status code, request status code description
```
mbucknell: Can we call this something more generic (maybe get_nwis_data). Since we pass in the url, this function's purpose is really to set the status code and status_reason to what we want for errors.

waterdata/views.py
```diff
@@ -5,9 +5,40 @@
...
+
+
+@app.route('/monitoring-location/<site_no>')
+def monitoring_location(site_no):
+    """
+    Monitoring Location view
+    :param site_no: USGS site number
+
+    """
+    rdb_resp_data = get_water_services_data(water_services, 'nwis/site/?site={}'.format(site_no))
```
mbucknell: Consider setting format=rdb as a query parameter. We are expecting rdb, maybe we should make it explicit.

waterdata/templates/monitoring_location.html
mbucknell: I'm also adding a title block to set the page's title as part of WDFN-26. I'm thinking we may want to put the station name in the <title> tag too.

jkreft-usgs: definitely should be using h1- and yes, the station name should be in the title block

waterdata/templates/errors/500.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Server Error</title>
+</head>
+<body>
+
+</body>
+</html>
```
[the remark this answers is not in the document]
ayan-usgs: I was under the impression from the slack discussion that sprucing up the errors pages was out of scope for this task...?

waterdata/templates/monitoring_location.html
ayan-usgs: Sure. The only reason h3 is being used is because that's the tag that's being using on the index. I don't really care one way or the other.

waterdata/templates/errors/500.html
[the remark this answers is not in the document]
mbucknell: It's fine to wait. Not sure if that future ticket had specific prettiness beyond the standard template.

config.py
```diff
@@ -1 +1,3 @@
 DEBUG = False
+
+WATER_SERVICES = 'https://waterservices.usgs.gov'
```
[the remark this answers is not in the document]
danielnaab: I'd vote for `SERVICE_ROOT = 'https://waterservices.usgs.gov/nwis/'` until we start using additional services.

waterdata/views.py
ayan-usgs: That's probably a @JanellFry-usgs question.

@JanellFry-usgs: What are all the possible 5XX that the NWIS services can reply with?

waterdata/tests/test_utils.py
ayan-usgs: It doesn't really matter but I seem to recall someone saying that we should ignore the more than one site coming back for this version of the view. Considering that there's a separate ticket to deal with that non-unique site codes.

config.py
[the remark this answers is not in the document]
mbucknell: I like Dan's suggestion. You could also potentially add SITE_ENDPOINT= SERVICE_ROOT/sites.

requirements.txt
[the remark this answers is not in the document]
danielnaab: I hear the IV/DV services may be the only ones supporting JSON. Something about performance problems with it...

Agree that Pandas is a big dependency to pull in. This code is being used reliably in the retriever, but could be refactored to be more "Pythonic", and to return a dictionary instead of a tuple: https://code.chs.usgs.gov/dnaab/nwisweb-retriever/blob/cmdline/retriever/lib/rdb.py  Maybe a namedtuple would be an efficient data structure to unpack into, rather than a dictionary.

requirements.txt
```diff
@@ -1,6 +1,17 @@
...
+python-dateutil==2.6.1
+pytz==2017.3
+requests==2.18.4
+requests-mock==1.4.0
+six==1.11.0
+urllib3==1.22
+Werkzeug==0.13
```
danielnaab: What do you all think about using Pipfile instead of requirements.txt? https://github.com/pypa/pipfile

I appreciate that all dependencies are pinned in this requirements.txt, but it is nice to keep the "first order" dependencies separate from the full list. Pipfile has a notion of a lock file, similar to how npm now works with package-lock.json.

waterdata/templates/errors/500.html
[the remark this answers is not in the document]
jkreft-usgs: no, just the standard template, rather than the completely unformatted 404 page that flask sends by default

waterdata/tests/test_utils.py
mbucknell: It's more a matter of testing a utility function. The function itself should parse any rdb file regardless of whether it's site information, data, or how many rows of data there are.

It should be ignored in terms of not handling it properly in the view.

requirements.txt
mbucknell: I'm leery of using it until it becomes stable.

waterdata/utils.py
```diff
@@ -0,0 +1,49 @@
+from io import StringIO
+from urllib.parse import urljoin
+
+import pandas as pd
+import requests as r
+
+
+def get_water_services_data(hostname, url_path):
```
jkreft-usgs: Since we are using requests, it might make sense to pass the path in as path along with a params dictionary

waterdata/tests/test_utils.py
ayan-usgs: I'm not arguing about the utility of tes,t I'm just pointing out what seems to be lapse in the argument about adding that test. Specifically, there are instances where we haven't over engineered something because it wasn't something we needed at the time if implementation. I would say that:

> Should also have a test that has more than one data row

Within the scope of this ticket is contrary to the arguments where were made during planning.

waterdata/views.py
danielnaab: Refer to here to see all response codes the site service sends: https://internal.cida.usgs.gov/stash/projects/NWISWEB/repos/site-service/browse/src/main/java/gov/usgs/service/si/SIServlet.java#292-319

I like passing all the 400s on, but based on the logic discussed in the Slack channel I feel 503 would be best for everything else.

waterdata/tests/test_utils.py
mbucknell: I think this is a more of a testing philosophy issue. A utility function may do more than what you are currently using it for in the code base. But if it was written to handle (in this case), any rdb file, than that's what should be tested, IMO.

ayan-usgs: I suppose one could argue that `parse_rdb` is an unnecessary expansion of scope for what this task  is expected to accomplish.

waterdata/utils.py
ayan-usgs: That function is generic enough that I'm considering just calling it `execute_get_request`.

requirements.txt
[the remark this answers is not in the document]
ayan-usgs: @danielnaab -- I've only used namedtuples when I felt like I needed immutability. I'm not sure they offer much of an advantage over a dictionary in this case.

waterdata/views.py
ayan-usgs: @danielnaab -- Agreed.

PR #16, opened 2018-01-02
Wdfn 26

mbucknell:
I modeled the USGS header on the one on the USGS visual ID page. For small screen sizes, I made the logo smaller and eliminated the USGS Home link since clicking on the logo already  provides that link


assets/src/styles/header.scss
```diff
@@ -0,0 +1,48 @@
...
+  height:1px;
+  overflow:hidden;
+}
+
+#skip a:focus
+{
+  position:static;
+  width:auto;
+  height:auto;
+}
```
danielnaab: Using SASS nesting for `#skip a` would be nice.

assets/src/styles/header.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+  .usgs-header-links-container {
+    font-size: $small-font-size;
+    @media screen and (max-width: $small-screen) {
+      #usgs-home-link {
+        font-size: 0px
+      }
+    }
+    @media screen and (min-width: $small-screen) {
```
danielnaab: Minor, but looks like we have a mixin we can use here: https://github.com/18F/web-design-standards/blob/develop/src/stylesheets/core/_utilities.scss#L46-L50

assets/src/styles/header.scss
```diff
@@ -0,0 +1,48 @@
+#skip a {
```
danielnaab: As an inclusion file / partial, I think this file should be named `_header.scss`

waterdata/templates/partials/header.html
```diff
@@ -1,57 +1,13 @@
...
-                </li>
-                <li>
-                    <a class="usa-nav-link" href="javascript:void(0)">
-                    <span>Distinct from each other</span>
-                    </a>
-                </li>
+        <nav class="usgs-header-links-container" role="navigation">
+            <ul>
+                <li aria-hidden="true" id="usgs-home-link"><a href="https://www.usgs.gov">USGS Home</a></li>
+                <li><a href="https://answer.usgs.gov/cgi-bin/gsanswers?tmplt=2">Contact USGS</a></li>
```
danielnaab: Should this be https://answers.usgs.gov?

mbucknell: Thanks. I thought I cut and pasted this correctly but it failed me again!

PR #21, opened 2018-01-04
Wdfn 14

JanellFry-usgs:


waterdata/constants.py
```diff
@@ -0,0 +1,48 @@
+"""
+Constants
+"""
+
+STATION_FIELDS_D = {
```
mbucknell: I wonder if we should put this in the config.py rather than a separate constants package. Any opinions?

danielnaab: It feels to me like something that should have it's own module.

mbucknell: One reason to put it in config.py would be that it's automatically available to the templates without passing as part of the context. I'm fine leaving as is though.

PR #22, opened 2018-01-05
Wdfn 19

JanellFry-usgs:


waterdata/views.py
```diff
@@ -17,20 +19,31 @@ def home():
...
-@app.route('/monitoring-location/<site_no>')
+@app.route('/monitoring-location/<site_no>', methods=['GET'])
 def monitoring_location(site_no):
     """
     Monitoring Location view
     :param site_no: USGS site number
 
     """
-    resp = execute_get_request(SERVICE_ROOT, path='/nwis/site/', params={'site': site_no, 'format': 'rdb'})
+    agency_cd = request.args.get('agency_cd')
```
mbucknell: Consider adding a test to make sure that the agency_cd is properly retrieved and used in execute_get_request. I think the most straightforward way to do this would be to use the unittest.mock library and patch 'utils.execute_get_request' and then verify that params is correctly set. Here is an example of a test that patches an imported function: https://github.com/USGS-CIDA/MLR-Ddot-Ingester/blob/master/tests/test_services.py#L35

PR #25, opened 2018-01-10
Wdfn 15 - Disambiguate Codes and Meanings for site service

mbucknell:
Includes the code used to generate the static json files. The static json files are in data/. The script is generate_lookup_json.py and uses the package loookup_generation.


lookup_generation/nwis_lookups.py
```diff
@@ -0,0 +1,64 @@
...
+    :param str code_key: Should not be null
+    :param str name_key: Should not be null
+    :param str desc_key: Can be the null string
+    :rtype: dict
+    """
+
+    def has_code_key(lookup):
+        return code_key in lookup
+
+    filtered_dict = filter(has_code_key, dict_iter)
```
ayan-usgs: I'm wondering if the code would be more concise if `has_code_key` is rewritten as a lambda function... :thinking:

lookup_generation/nwis_lookups.py
```diff
@@ -0,0 +1,64 @@
...
+    :return: dict
+    """
+
+    def has_code_key(lookup):
+        return code_key in lookup
+
+    def get_code(d):
+        return d.get(code_key)
+
+    filtered_dict = filter(has_code_key, dict_iter)
```
ayan-usgs: See comment above, though I wonder since `has_code_key` appears twice now, if it would make more sense to have it as it's own standalone function rather than as a lambda function or a nested one as it is now.

waterdata/utils.py
```diff
@@ -53,3 +53,55 @@ def parse_rdb(rdb_iter_lines):
...
+    :param dict country_state_county_lookups:
+    :rtype: dict
+    """
+
+    def get_state_name (country_code, state_code):
+        """
+        Return the name of the state with country_code and state_code
+        :param str country_code:
+        :param str state_code:
+        :rtype: str
```
ayan-usgs: It looks like this is actually returning a dictionary rather than a string.

lookup_generation/nwis_lookups.py
mbucknell: The two has_code_keys are different in that they use the passed in code_key variable. And I have to be honest, I'm not a big fan of lambda functions. I think a named function makes the code more readable and understandable. For instance the above function is clearly filter on code_key.

waterdata/utils.py
mbucknell: No it returns a string, at least it should be. It returns the state name only.

lookup_generation/nwis_lookups.py
ayan-usgs: Ah yes. I missed that about the `code_key`. Sorry.

waterdata/utils.py
ayan-usgs: Yep. You're right. I was looking at the returned value of the parent function. To honest, while nested functions are certainly useful, I find them difficult to read.

lookup_generation/nwis_lookups.py
danielnaab: To add a third perspective.. stylistically I find list and dictionary comprehensions clearer to read. Especially when dealing with nested data, you can discern the shape of the return value at a quick glance.

mbucknell: I like list comprehensions for mapping lists, but not as fond of them for filtering. So we have three distinct opinions! I think I'll keep it as is for now.

PR #29, opened 2018-01-11
Wdfn 16 json ld

ayan-usgs:


waterdata/location.py
```diff
@@ -0,0 +1,173 @@
...
+class MonitoringLocation:
+
+    service_endpoint = app.config['SERVICE_ROOT']
+
+    def __init__(self, monitoring_location_record):
+        """
+        A monitoring location object.
+
+        Attributes:
+            location_number (str): identifier for the monitoring location
```
ayan-usgs: I decided to retain as attribute names as I get the sense that we don't want to call these places sites in the future.

jkreft-usgs: Your sense is correct

waterdata/tests/test_location.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def setUp(self):
+        # an actual test record may include many more columns from an RDB file record
+        self.test_record = {'site_no': '01630500',
+                            'agency_cd': 'USGS',
+                            'station_nm': 'POTOMAC RIVER NEAR WASH, DC LITTLE FALLS PUMP STA',
+                            'dec_lat_va': '38.94977778',
+                            'dec_long_va': '-77.12763889'
+                            }
+        self.test_rdb_text = ('#\nagency_cd\tsite_no\tstation_nm\tsite_tp_cd\tdec_lat_va\tdec_long_va\tcoord_acy_cd\t'
```
danielnaab: Personal preference - but for readability I'd prefer a multi-line """xxx""" style string.

waterdata/location.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Determine the parameters measured at the site.
+
+        :return: USGS parameter codes measured at a site
+        :rtype: set
+
+        """
+        if (self._resp is None or self._site_param_info is None) or self._resp.status_code != 200:
+            self._resp, self._site_param_info = self._get_location_parameter_info()
+        supported_params = set([timeseries['parm_cd'] for timeseries in self._site_param_info])
```
danielnaab: You can pass an iterator to set here, and avoid creating an intermediate list: `set(timeseries['parm_cd'] for timeseries in self._site_param_info)`

waterdata/location.py
```diff
@@ -0,0 +1,173 @@
...
+
+        """
+        for key, value in monitoring_location_record.items():
+            setattr(self, '_{}'.format(key), value)
+        self.location_number = getattr(self, '_site_no')
+        self.agency_code = getattr(self, '_agency_cd')
+        self.location_name = getattr(self, '_station_nm')
+        self.latitude = getattr(self, '_dec_lat_va')
+        self.longitude = getattr(self, '_dec_long_va')
+        self._resp = None
```
danielnaab: I don't see this being used anywhere - do we need to keep it around?

waterdata/location.py
```diff
@@ -0,0 +1,173 @@
...
+        :param str parameter_cd: the USGS parameter code of interest
+        :return: a parameter's "start_date", "end_date", and "record_count" if available; dates are Python date objects
+        :rtype: collection.namedtuple or None
+
+        """
+        if (self._resp is None or self._site_param_info is None) or self._resp.status_code != 200:
+            self._resp, self._site_param_info = self._get_location_parameter_info()
+        try:
+            param_series = next((timeseries for timeseries in self._site_param_info
+                                 if timeseries['parm_cd'] == parameter_cd))
```
danielnaab: Nice trick :)

waterdata/location_utils.py
```diff
@@ -0,0 +1,148 @@
...
+                             if parameter_record['parm_cd'] == parameter_cd))
+    except StopIteration:
+        return None
+    else:
+        record_start_date = param_series['begin_date']
+        record_end_date = param_series['end_date']
+        start_date = datetime.datetime.strptime(record_start_date, '%Y-%m-%d').date()
+        end_date = datetime.datetime.strptime(record_end_date, '%Y-%m-%d').date()
+        record_count = param_series['count_nu']
+        parameter = Parameter(parameter_cd=parameter_cd,
```
mbucknell: Consider returning 'Parameter...' rather than declaring a local variable.

waterdata/location_utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def build_linked_data(location_number, location_name, agency_code, latitude, longitude, location_capabilities):
+    """
+    Given site metadata, construct a dictionary / json-ld for the site.
+    The constructed json-ld conforms to the context documents
+    at https://opengeospatial.github.io/ELFIE/json-ld/elf-index.jsonld
+    and https://opengeospatial.github.io/ELFIE/json-ld/hyf.jsonld.
+
+    If a site is missing a piece of data that is needed by the json-ld,
```
mbucknell: I find this statement misleading. An empty dictionary is never returned.

waterdata/templates/base.html
```diff
@@ -8,6 +8,7 @@
         <link rel="stylesheet" href="{{ config.STATIC_ROOT }}/main.css">
         <script type="text/javascript">var SERVICE_ROOT = '{{ config.SERVICE_ROOT }}/nwis';</script>
         <script async src="{{ config.STATIC_ROOT }}/bundle.js"></script>
+        {% block linked_data %}{% endblock %}
```
mbucknell: Consider making this a more general tag ... maybe something like {% block page_script %}. Then its clear it could be used anything really... page specific javasscript, json ld, etc.

waterdata/tests/test_location.py
```diff
@@ -0,0 +1,371 @@
...
+                            'count_nu': '198'
+                            }
+        self.param_00065 = {'parm_cd': '00065',
+                            'begin_date': '2007-10-01',
+                            'end_date': '2018-01-10',
+                            'count_nu': '800'
+                            }
+        self.test_rdb_param_data = [self.param_00010, self.param_00060, self.param_00065, self.param_00095]
+
+    def test_get_capabilities(self):
```
mbucknell: Consider adding a test for passing in an empty iterable.

waterdata/location_utils.py
```diff
@@ -0,0 +1,148 @@
...
+def get_site_parameter(location_parameter_records, parameter_cd):
+    """
+    Determine the period of record and number of records for a
+    parameter being measured at this site. If the parameter code
+    value specified is not available at this site, None is returned.
+
+    :param iterable location_parameter_records: an iterable containing of location parameter
+    :param str parameter_cd: the USGS parameter code of interest
+    :return: a parameter's "start_date", "end_date", and "record_count" if available; dates are Python date objects
+    :rtype: collection.namedtuple or None
```
mbucknell: I wonder if there is a standard way to say we are returning a Parameter namedtuple?

waterdata/tests/test_views.py
```diff
@@ -46,25 +44,83 @@ def setUp(self):
...
+        m_resp_param.status_code = 200
+        m_resp_param.iter_lines.return_value = iter(self.parameter_lines)
+
+        return_values = [m_resp, m_resp_param]
+        r_mock.side_effect = return_values
+
+        response = self.app_client.get('/monitoring-location/{}?agency_cd=USGS'.format(self.test_site_number))
         self.assertEqual(response.status_code, 200)
         self.assertIn('Some Random Site', response.data.decode('utf-8'))
+        self.assertIn('@context', response.data.decode('utf-8'))
```
mbucknell: This is just something I've been think about as far as testing views. What do you think of mocking the call to render_template rather than check the contents of the template? It would remove the dependency on the template. Just looking for your opinion.

waterdata/views.py
```diff
@@ -53,6 +76,10 @@ def monitoring_location(site_no):
             'STATION_FIELDS_D'  : STATION_FIELDS_D
         }
         http_code = 200
+        # don't want to create more DOM elements if we don't have to
```
mbucknell: Consider just using {% if json_ld %} in the template, then the if statement can be eliminated (see http://jinja.pocoo.org/docs/2.10/templates/#if)

waterdata/location_utils.py
ayan-usgs: Yeah. That wasn't updated when the class was removed.

waterdata/location_utils.py
```diff
@@ -0,0 +1,148 @@
...
+        transformed_location[key] = transformed_value
+
+    return transformed_location
+
+
+def get_capabilities(location_parameter_records):
+    """
+    Determine the parameters measured at the site.
+
+    :param iterable location_parameter_records: an iterable containing of location parameters
```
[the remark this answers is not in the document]
ayan-usgs: Sure.

waterdata/location_utils.py
```diff
@@ -0,0 +1,148 @@
...
+    value specified is not available at this site, None is returned.
+
+    :param iterable location_parameter_records: an iterable containing of location parameter
+    :param str parameter_cd: the USGS parameter code of interest
+    :return: a parameter's "start_date", "end_date", and "record_count" if available; dates are Python date objects
+    :rtype: collection.namedtuple or None
+
+    """
+    try:
+        param_series = next((parameter_record for parameter_record in location_parameter_records
```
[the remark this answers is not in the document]
ayan-usgs: I think I'll leave it as it is. I'm don't see that as simplifying the code very much.

waterdata/location_utils.py
ayan-usgs: I haven't been able to find a standard or best practice on it's own. I've been wondering if it would make more sense to document it like a class in it's own right (e.g. :rtype: waterdata.location_utils.Parameter or None). In many ways, namedtuples can be thought as something one would use in-lieu of creating a normal Python class.

waterdata/tests/test_views.py
ayan-usgs: From a unit test perspective, I think it makes sense because then we're just testing the view code and making sure that `render_template` is being called with the parameters we expect.

On the other-hand, there it's nice to also be able to test the logic in the jinja templates. Perhaps what that really get's at is that we should try to unit test Jinja templates as discussed here.

So, I'd lean slightly towards mocking `render_template` and writing up distinct tests for anything we might want to test in a template.

waterdata/location_utils.py
mbucknell: I like that!

PR #36, opened 2018-01-16
Add interactive grapher link to locations page.

danielnaab:


config.py
```diff
@@ -9,7 +9,7 @@
 DEBUG = False
 
 SERVICE_ROOT = 'https://waterservices.usgs.gov'
-LEGACY_ENDPOINTS = {
+CGI_ENDPOINTS = {
```
mbucknell: I don't like calling these CGI_ENDPOINTS. Maybe we could say NWIS_ENDPOINTS?

danielnaab: Sure

PR #40, opened 2018-01-17
Wdfn 18

JanellFry-usgs:


waterdata/templates/monitoring_location.html
```diff
@@ -13,32 +13,61 @@
 {% block content %}
     <div class="usa-width-one-whole">
         {% if status_code == 200 %}
-            <h1>Station Name: {{ station.station_nm }}</h1>
+
+            {% if station|length > 1 %}
+
+            <h1>Multiple sites found</h1>
```
mbucknell: Please indent within an {% if %} and the {% for block %}

waterdata/views.py
```diff
@@ -33,51 +33,77 @@ def monitoring_location(site_no):
...
-                app.config['NWIS_CODE_LOOKUP'],
-                app.config['COUNTRY_STATE_COUNTY_LOOKUP']
-            ),
-            'STATION_FIELDS_D'  : STATION_FIELDS_D,
-            'json_ld': safe_json_ld
-        }
+
+        station_nm = ''
+        site_no = ''
+        if len(data_list) == 1:
```
mbucknell: This should be reworked so that this if clause is only made once (they other is on line 63).

waterdata/templates/monitoring_location.html
```diff
@@ -13,32 +13,61 @@
...
                     <th scope="col">Description</th>
                     <th scope="col">Station Data</th>
                 </tr>
             </thead>
+
             <tbody>
-                {% for key, value in location_with_values.items() %}
-                <tr>
-                    <th scope="row"> {{ STATION_FIELDS_D[key] }} </th>
+                 {% for key, value in location_with_values.items() %}
```
mbucknell: It looks like this is using 5 spaces for indentation. It should be four.

waterdata/views.py
```diff
@@ -37,47 +37,72 @@ def monitoring_location(site_no):
...
+                   'station': data_list,
+                   'station_nm': station_nm,
+                   'site_no': site_no,
+                   'STATION_FIELDS_D': STATION_FIELDS_D}
+
+        if len(data_list) == 1:
+            station_nm = data_list[0]['station_nm']
+            site_no = data_list[0]['site_no']
+
+            station_record = data_list[0]
```
danielnaab: This is a minor thing - but I think it would be slightly clearer if `data_list` was named `stations` then `station = stations[0]` in this if condition. You could also get rid of `station_nm` and `site_no`, and reference them off the dictionary like the other fields.

PR #43, opened 2018-01-18
WDFN-30 - make the seven-day hydrograph accessable

mbucknell:


package.json
```diff
@@ -41,6 +41,7 @@
     "babel-preset-env": "^1.6.1",
     "babelify": "^8.0.0",
     "browserify": "^15.1.0",
+    "eslint": "^4.15.0",
```
mbucknell: I added this to the dev dependencies because PyCharm seems to want it to be in the node_modules directory in order for ESLINTing to work in the IDE.

assets/src/scripts/hydrograph/axes.js
```diff
@@ -39,13 +39,13 @@ function createAxes(xScale, yScale, yTickSize) {
...
- * @param  {Object} options.yLabelLoc {x, y} location of y-axis label
- * @param  {String} options.yTitle    y-axis label
+ * @param  {Object} plot      Node to append to
+ * @param  {Object} xAxis     D3 Axis x-axis
+ * @param  {Object} yAxis     D3 Axis y-axis
+ * @param  {Object} xLoc      {x, y} location of x-axis
+ * @param  {Object} yLoc      {x, y} location of y-axis
+ * @param  {Object} yLabelLoc {x, y} location of y-axis label
+ * @param  {String} yTitle    y-axis label
  */
```
danielnaab: The options.xxx syntax indicates that the parameter is a single object, by convention called options. Here's the JSDoc syntax: http://usejsdoc.org/tags-param.html#parameters-with-properties

assets/src/scripts/hydrograph/index.spec.js
```diff
@@ -149,4 +179,4 @@ const MOCK_DATA = `[
         "time": "2018-01-03T18:15:00.000Z",
         "value": 24
     }
-]`
+];
```
danielnaab: Hmmm... why did my code work... it shouldn't have? I must've originally intended for MOCK_DATA to be a json string to be parsed...

assets/src/scripts/index.js
```diff
@@ -1,19 +1,24 @@
...
     let nodes = document.getElementsByClassName('hydrograph');
     for (let node of nodes) {
         getTimeseries({sites: [node.dataset.siteno]}, series => {
-            let dataIsValid = series[0] && !series[0].values.some(d => d.value == -999999);
+            let dataIsValid = series[0] && !series[0].values.some(d => d.value === -999999);
             new Hydrograph({
                 element: node,
                 data: dataIsValid ? series[0].values : [],
-                title: dataIsValid ? series[0].description : 'No data'
+                yLabel: dataIsValid ? series[0].description : 'No data',
```
danielnaab: Somehow, running locally - the y axis label is now the default "Data" rather than the timeseries description.

assets/src/scripts/models.js
```diff
@@ -34,8 +35,8 @@ function get(url, callback) {
 
 /**
  * Get a given timeseries dataset from Water Services.
- * @param  {Array}    options.sites  Array of site IDs to retrieve.
- * @param  {Array}    options.params List of parameter codes
```
danielnaab: Ditto from above - I think we want to keep options here

assets/src/scripts/accessibility.js
```diff
@@ -0,0 +1,60 @@
...
+        .enter().append('td')
+            .attr('scope', 'row')
+            .text(function(d) {
+                return d;
+            });
+
+}
+
+module.exports = {addSVGAccessibility, addSROnlyTable};
+
```
danielnaab: I trust this is all good WRT screen-readability - but maybe we could add a link to some docs describing standards (or best practices)? Would be useful to read up on when changes are made.

assets/src/scripts/hydrograph/index.spec.js
mbucknell: I think because the test only tests to see if the path is there, the path is still generated even if the time series is empty.

assets/src/scripts/index.js
mbucknell: Thanks, I see that now. I'll fix

assets/src/scripts/models.js
mbucknell: Agreed. It seemed like there were inconsistancies but it was probably inconsistancies that I introduced 😄

PR #46, opened 2018-01-22
WDFN-35 Site metadata updates

danielnaab:


waterdata/location_utils.py
```diff
@@ -6,11 +6,13 @@
...
 import datetime
 
+from flask import url_for
+
 
 Parameter = namedtuple('Parameter', ['parameter_cd', 'start_date', 'end_date', 'record_count'])
 
 
-def get_disambiguated_values(location, code_lookups, country_state_county_lookups):
+def get_disambiguated_values(location, code_lookups, country_state_county_lookups, huc_lookups):
```
danielnaab: I decided to not try to do this now - but breaking this function into smaller pieces might be advantageous. As we add more lookups, the complexity goes up, and all the existing tests need to be modified.

ayan-usgs: Sounds like a plan. :+1:

PR #47, opened 2018-01-22
WDFN-66 - Switch to linear scale and add regression test

danielnaab:


assets/src/scripts/hydrograph/scales.spec.js
```diff
@@ -34,4 +34,10 @@ describe('Charting scales', () => {
         expect(range[0]).toEqual(100);
         expect(range[1]).toEqual(0);
     });
+
+    fit('WDFN-66: yScale deals with range [0,1] correctly', () => {
```
mbucknell: I think this should be excluded? But shouldn't be a focused test

danielnaab: Oops - yep that that I know that trick I'm going wild.

PR #49, opened 2018-01-23
Mapping WDFN-49

danielnaab:


assets/src/scripts/index.js
```diff
@@ -1,29 +1,21 @@
...
 
-const { getTimeseries } = require('./models');
-const Hydrograph = require('./hydrograph');
 
-// Create a time formatting function from D3's timeFormat
-const formatTime = timeFormat('%c %Z');
+const COMPONENTS = {
+    hydrograph: require('./components/hydrograph').attachToNode,
+    map: require('./components/map').attachToNode
+};
```
danielnaab: Each component would have a line here

assets/src/scripts/index.js
```diff
@@ -1,29 +1,21 @@
...
-            let dataIsValid = series[0] && !series[0].values.some(d => d.value === -999999);
-            new Hydrograph({
-                element: node,
-                data: dataIsValid ? series[0].values : [],
-                yLabel: dataIsValid ? series[0].variableDescription : 'No data',
-                title: dataIsValid ? series[0].variableName : '',
-                desc: dataIsValid ? series[0].variableDescription + ' from ' + formatTime(series[0].seriesStartDate) + ' to ' + formatTime(series[0].seriesEndDate) : ''
-            });
-        });
+        COMPONENTS[node.dataset.component](node, node.dataset);
```
danielnaab: Components attached to the DOM node here - data attributes passed as component options.

assets/src/styles/main.scss
```diff
@@ -5,4 +5,11 @@ $usgs-blue: #02274B;
...
 
-@import './hydrograph';
+.wdfn-component {
+    &[data-component='hydrograph'] {
+        @import './components/hydrograph';
+    }
+    &[data-component='map'] {
+        @import './components/map';
+    }
+}
```
danielnaab: Scopes the CSS per-component in a consistent way

waterdata/templates/monitoring_location.html
```diff
@@ -6,6 +6,11 @@
         {{ json_ld }}
     </script>
     {% endif %}
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.3.1/dist/leaflet.css" integrity="sha512-Rksm5RenBEKSKFjgI3a41vrjkw4EVPlJ3+OiI65vTjIdo9brlAacEuKOiQ5OFh7cOI1bkDwLqdLw3Zg0cRJAAQ=="crossorigin=""/>
+    <script src="https://unpkg.com/leaflet@1.3.1/dist/leaflet.js" integrity="sha512-/Nsx9X4HebavoBvEBuyp3I7od5tA0UzAxs+j83KgC8PU0kgB4XiK4Lfe4y4cgBtaRJQEIFCW+oC506aPT2L1zw==" crossorigin=""></script>
+    <script src="https://unpkg.com/esri-leaflet@2.1.2/dist/esri-leaflet.js"
+    integrity="sha512-ouokQ1RIIoqPTZKwlapdxHO5VWFoAi8wE+SwhSX89Ifac0w3p+H2da4oqHvRsBTIpNLewzAZU7gRVDFXyXcfjA=="
+    crossorigin=""></script>
```
danielnaab: This is definitely debatable - rather than pulling in via npm, this is using a Cloudflare hosted CDN for Leaflet 1.3.1. If we ditch this approach, perhaps we should create a separate app bundle for the Leaflet JS, so we can only load it on the pages where it's needed.

An option, also, would be for the component's initialization code to use something similar to jQuery's $.getScript() function, so the markup doesn't need to track these dependencies.

assets/src/scripts/components/map.js
```diff
@@ -0,0 +1,22 @@
+const HYDRO_URL = 'https://tiles.arcgis.com/tiles/P3ePLMYs2RVChkJx/arcgis/rest/services/Esri_Hydro_Reference_Overlay/MapServer';
+
+
+function attachToNode(node, {latitude, longitude, zoom}) {
```
mbucknell: There will be cases (maybe @jkreft-usgs  can find one for you) where latitude/longitude will be the null. In those cases, I'm assuming we don't display the map.

waterdata/templates/monitoring_location.html
mbucknell: I think soon we should probably think about having different bundles for different pages. Is this usually done? Or I don't see the difference between Leaflet and D3 in terms of there are pages in the application that don't use D3 either. I think we should be consistent, so I would prefer that these be brought in with npm.

waterdata/templates/monitoring_location.html
```diff
@@ -20,7 +25,10 @@ <h1>Station Name: {{ station.station_nm }}</h1>
                 <li><a href="{{ config.NWIS_ENDPOINTS.UV }}/?cb_00060=on&format=gif_mult_parms&site_no={{ station.site_no }}&period=7" target="_blank">Interactive grapher</a></li>
                 <li><a href="{{ config.NWIS_ENDPOINTS.UV }}/?format=img_default&site_no={{ station.site_no }}&period=7" target="_blank">Presentation Charts</a></li>
             </ul>
-            <div class="hydrograph" data-siteno="{{ station.site_no }}"></div>
+
+            <div class="wdfn-component" data-component="hydrograph" data-siteno="{{ station.site_no }}"></div>
+            <div class="wdfn-component" data-component="map" data-latitude="{{ station.dec_lat_va }}" data-longitude="{{ station.dec_long_va }}" data-zoom=8></div>
```
mbucknell: I like this a lot!

waterdata/templates/monitoring_location.html
danielnaab: Yes we could do that. We'd want to have a common bundle for D3, etc, so they don't get downloaded multiple times. Possibly also with some lazy loading.

PR #60, opened 2018-01-25
WDFN-55 - Social sharing links

danielnaab:


assets/src/scripts/accessibility.js
```diff
@@ -49,7 +49,7 @@ function addSROnlyTable({container, columnNames, data}) {
         .data(function(d) {
             return d;
         })
-        .enter().append('td')
+        .enter().append('th')
```
mbucknell: This should not have been changed. It's selecting on td

assets/src/styles/main.scss
```diff
@@ -1,12 +1,14 @@
...
 
 @import './uswds';
+@import '../../../node_modules/font-awesome/scss/font-awesome';
 
 $usgs-blue: #02274B;
 
-@import './tooltip';
-@import './header';
-@import './footer';
+@import './partials/footer';
```
mbucknell: Do we really need a separate directory for the partials? Seems redundant with the file naming with an underscore.

package.json
```diff
@@ -63,6 +63,7 @@
   "dependencies": {
     "d3": "^4.12.2",
     "esri-leaflet": "^2.1.2",
+    "font-awesome": "^4.7.0",
```
mbucknell: I do like font awesome!

assets/src/scripts/accessibility.js
danielnaab: The aXe accessibility validator caught an issue there - that the scope attribute isn't defined on `td`. I suppose the select should be updated also?

mbucknell: yes.

assets/src/styles/main.scss
danielnaab: Not strictly needed, of course... just seemed useful to distinguish pages/component/partials in that manner. I'm inclined to keep it like that but don't feel too strongly about it.

mbucknell: I also don't feel strongly about this.

assets/src/scripts/accessibility.js
danielnaab: Maybe it would be more appropriate to remove the scope attribute, since it's not a label for the row.

mbucknell: I would be okay with that for now. Jim and I were just talking about how to add the previous year's time series data, whether it should be a separate table or in this table. If we included it in this table, I think it might be appropriate to label the cell containing the date with scope=row. When the second line gets added each <tr> would have two <th> elements one for this year's data and one for last years both with scope=row which is allowed.

danielnaab: Ah yes, that makes sense.

PR #61, opened 2018-01-25
Wdfn 54 - Adding the ability to show last year's data on the hydrograph

mbucknell:
Known deficiencies: 
* Need to add tests for getPreviousYearTimeseries in models
* The Voronoi has not been added.  We may want to move that work to a new ticket
* Need to add a config.py variable for the iv service to use when data is older than 120 days.

Nice to do: add tests for ajax.js

Things to note: 
* Using a new plugin proxyquireify to proxy module functions within the tests. Using in models.spec.js to mock the ajax.get call.
* models.js has been reworked to use the new ajax.get.
* Included the new checkbox within the hydrograph component


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -43,23 +43,74 @@ class Hydrograph {
...
+
+        // Add the new time series
+        this._plotDataLine(this.plot, {xScale: xScale, yScale: this.scale.yScale}, 'compare');
+    }
+
+    /**
+     * Remove the compare time series from the plot and rescale
+     */
+    removeCompareTimeSeries() {
+        // Remote the compare time series
```
ayan-usgs: Nit-picky thing -- should that be "Remove the..."?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -77,34 +128,34 @@ class Hydrograph {
...
-        const {xAxis, yAxis} = createAxes(xScale, yScale, -WIDTH + MARGIN.right);
+        this.axis = createAxes(this.scale.xScale, this.scale.yScale, -WIDTH + MARGIN.right);
 
         // Draw the graph components with the given scaling.
         appendAxes({
-            plot,
-            xAxis,
-            yAxis,
+            plot: this.plot,
+            xAxis: this.axis.xAxis,
```
ayan-usgs: Still being a D3 neophyte, what's happening to the tick marks on the x-axis? Do they still so labels like "Jan 21, 2018"? Since the `createAxes` function doesn't appear to have changed, it looks like it might. Would that be confusing to a user comparing data with 2017, or would the legend be sufficent?

danielnaab:
Alternative to using voronoi for tooltip detection - the way the TX WSC does it seems to work well: https://txpub.usgs.gov/dss/gwis/0.0/doc/samples/sample03_code.html

assets/src/scripts/components/hydrograph/index.js
mbucknell: I don't change the x axis and yes it still shows Jan 21, 2018. I think I will change the formatting to not show the year. The two time series are plotted using different xScales. And that's another thing that needs to be done is adding a legend but I think that should be a different ticket at this point.

danielnaab: I've vote for having an separate axis label for each date range - could color-code the labels rather than add a legend.

danielnaab: One bonus of separate labels is it will continue to work if we do "compare to last week" or similar.

mbucknell: So visually would you suggest stacking the labels or putting an axis at the top?

mbucknell: An axis on top would be my preference

danielnaab: Not sure. I feel like I need to pick up a book on data visualization.

mbucknell: I also think adding a legend with year would work if we are always comparing year to year.  But I agree with Dan's point which would also make handling leap years clearer.  So still leaning toward two axes. I'll try coding that up to see how it looks.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -43,23 +43,75 @@ class Hydrograph {
...
+        updateYScale(this.scale.yScale, currentYExtent);
+        updateYAxis(this.axis.yAxis, this.scale.yScale);
+        this.svg.select('.y-axis')
+            .call(this.axis.yAxis);
+
+        //Redraw the current ts
+        select('#ts-current')
+            .attr('d', this.currentLine(this._tsData.current));
+    }
+
```
danielnaab: From some reading I've been doing, I think there's a pattern we can use for adding/removing using `selection.enter().append()` to add new elements, them `selection.exit().remove()` to dispose of the old ones.

I think your code looks good as-is, though, it's more of a future thing.

mbucknell: Seems like enter() and exit() is useful if the data changed but maybe it can be used to add/remove new things to the graph. Will be interesting to explore.

PR #65, opened 2018-01-30
Wdfn 51 median statistic graph

ayan-usgs:


assets/src/scripts/models.js
```diff
@@ -42,33 +42,126 @@ export function getTimeseries({sites, params=['00060'], startDate=null, endDate=
...
+
+/**
+ * Determine if a given year is a leap year
+ *
+ * @param year
+ * @returns {boolean}
+ */
+export function isLeapYear(year) {
+    return (year % 4) === 0;
+}
```
danielnaab: There are some special rules concerning years divisible by 100 and 400, also. https://www.wwu.edu/skywise/leapyear.html

assets/src/scripts/models.js
```diff
@@ -42,33 +42,126 @@ export function getTimeseries({sites, params=['00060'], startDate=null, endDate=
...
+
+/**
+ *  Read median RDB data into something that makes sense
+ *
+ * @param medianData
+ * @param timeSeries
+ * @params subsetDays
+ * @returns {Array}
+ */
+export function parseMedianData(medianData, timeSeries) {
```
danielnaab: I wonder if this function would be easier to understand if, rather than taking in `timeSeries`, it took in only the derived values that are needed - like start and end date?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -107,6 +115,10 @@ class Hydrograph {
         //Redraw the current ts
         select('#ts-current')
             .attr('d', this.currentLine(this._tsData.current));
+        //remove the median stats plot and recreate
+        this.svg.selectAll('#median-point').remove();
+        this.svg.selectAll('#median-text').remove();
+        this._plotMedianPoints();
```
danielnaab: When I merge this into my current branch, I'll have to modify this a bit - but I was wondering why we need the refresh here as well as in the constructor?

assets/src/scripts/models.js
```diff
@@ -42,33 +42,126 @@ export function getTimeseries({sites, params=['00060'], startDate=null, endDate=
...
+    let url = `${SERVICE_ROOT}/stat/?format=rdb&sites=${sites.join(',')}&statReportType=daily&statTypeCd=${statType}&parameterCd=${params.join(',')}`;
+    return get(url);
+}
+
+
+/**
+ * Function to parse RDB to Objects
+ */
+export function parseRDB(rdbData) {
+    let rdbLines = rdbData.split('\n');
```
danielnaab: I suppose there's not a great way to do this in the browser, without pulling in a new dependency - but it might be worth researching if we can iterate over this without splitting and creating the intermediate array.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: The refresh happens here so that when the compare-timeseries is removed and the y-axis rescales, the median points will be re-positioned according to the updated y-axis. I was at a loss as to how to get that to happen more elegantly. :thinking:

assets/src/scripts/models.js
ayan-usgs: I'll take a look into that.

assets/src/scripts/models.js
ayan-usgs: Huh... wasn't aware of all that. Thanks!

assets/src/scripts/components/hydrograph/index.js
mbucknell: You might want to look at using the .call function. https://github.com/d3/d3-selection#selection_call

ayan-usgs: Thanks for the suggestion.

PR #66, opened 2018-01-31
revert to removing median points and recreating them... using .call()…

ayan-usgs:
… wasn't working out

### DIFF
--- a/assets/src/scripts/components/hydrograph/index.js
+++ b/assets/src/scripts/components/hydrograph/index.js
@@ -87,8 +87,10 @@ class Hydrograph {
         select('#ts-current')
             .attr('d', this.currentLine(this._tsData.current));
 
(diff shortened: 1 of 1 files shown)


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -87,8 +87,10 @@ class Hydrograph {
         select('#ts-current')
             .attr('d', this.currentLine(this._tsData.current));
 
-        //redraw the median stats scatter plot
-        this.svg.selectAll('#median-point').call(this._plotMedianPoints, this.scale, this._medianStatsData);
```
ayan-usgs: Doing this was the page to freeze up after a few clicks of the "compare" checkbox and wasn't successfully redrawing the median statistic points when the y-axis updates.

PR #68, opened 2018-02-01
Wdfn 77 hydrograph legend -- WIP (don't merge)

ayan-usgs:


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -238,6 +235,126 @@ class Hydrograph {
...
+                let detachedMarkerBBox = detachedMarker.node().getBBox();
+                legendGroup.append('text')
+                    .attr('x', detachedMarkerBBox.x + detachedMarkerBBox.width + markerTextOffset)
+                    .attr('y', 0)
+                    .text(legendMarker.text);
+            }
+            previousMarkerGroup = legendGroup;
+        }
+
+        // let legendGroup1 = legend.append('g');
```
ayan-usgs: Commented out code from "manual" creation of the legend.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -238,6 +233,126 @@ class Hydrograph {
             });
     }
 
+    _plotLegend() {
```
ayan-usgs: Is all this more complicated than it needs to be?

mbucknell: Seems like a reasonable approach to me. I didn't do a full review because it seems like you are still working on it.

assets/src/scripts/components/hydrograph/index.spec.js
```diff
@@ -117,7 +118,48 @@ describe('Hydrograph charting module', () => {
...
+                    {
+                        type: lineMarker,
+                        domId: null,
+                        domClass: null,
+                        text: 'blah',
+                        groupId: null
+                    },
+                    {
+                        type: circleMarker,
+                        r: 4,
```
mbucknell: Should we use CIRCLE_RADIUS here, too?

assets/src/scripts/components/hydrograph/store.js
```diff
@@ -129,20 +146,80 @@ export const timeSeriesReducer = function (state={}, action) {
...
+                    beginYear = state.statisticalMetaData.beginYear;
+                    endYear = state.statisticalMetaData.endYear;
+                    text = `Median Discharge ${beginYear} - ${endYear}`;
+                }
+                catch(err) {
+                    beginYear = '';
+                    endYear = '';
+                    text = 'Median Discharge';
+                }
+                marker = defineCircleMarker(4, null, 'median-data-series', text, 'median-circle-marker');
```
mbucknell: Shoulde we use CIRCLE_RADIUS here?

assets/src/scripts/components/hydrograph/points.js
```diff
@@ -7,12 +7,11 @@ const { createSelector, defaultMemoize: memoize } = require('reselect');
  * @param  {String} tsDataKey Timeseries key
  * @return {Array}            Array of points.
  */
-const pointsSelector = memoize(tsDataKey => createSelector(
+const pointsSelector = memoize((tsDataKey) => createSelector(
```
ayan-usgs: Darn, I thought I removed those extra parens...

assets/src/scripts/components/hydrograph/store.js
```diff
@@ -79,6 +77,18 @@ export const Actions = {
...
             medianStatistics
         };
+    },
+    setLegendMarkers(key) {
+        return {
+            type: 'SET_LEGEND_MARKERS',
+            key
+        };
+    },
+    selectLegendMarkers() {
```
mbucknell: I wonder if this should be called something like showLegendMarkers and have the type "Show_Legend_Markers? I'm getting a little confused with the names "display_markers" and "legend_markers"? I'm not sure the name I suggested is better.

assets/src/scripts/components/hydrograph/markers.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    describe('circleMarker', () => {
+        let r = 10;
+        let x = 2;
+        let y = 2;
+
+        it('Returns circle element with correct attributes', () => {
+            let circle = circleMarker({r: r, x: x, y: y});
+            let node = circle.node();
```
ayan-usgs: :man_facepalming:

assets/src/scripts/components/hydrograph/store.js
```diff
@@ -129,20 +140,80 @@ export const timeSeriesReducer = function (state={}, action) {
...
+                    ...state.legendMarkers,
+                    [action.key]: marker
                 }
             };
 
+        case 'SELECT_DISPLAY_MARKERS':
+            let displayMarkers = [];
+            const showSeries = state.showSeries ? state.showSeries : {};
+            const markers = state.legendMarkers ? state.legendMarkers : {};
+            for (const [key, value] of Object.entries(showSeries)) {
```
danielnaab: I'm not 100% certain, but I don't think we have a polyfill for `Object.entries` with our build tooling.

assets/src/scripts/components/hydrograph/store.js
```diff
@@ -129,20 +140,80 @@ export const timeSeriesReducer = function (state={}, action) {
...
+                    if (marker) {
+                        displayMarkers.push(marker);
+                    }
+                }
+            }
+            return {
+                ...state,
+                displayMarkers: displayMarkers
+            };
+
```
danielnaab: I'm wondering if rather than save the markers in the state, we should have a selector that calculates them based on what's currently visible. In an idealized situation, the reducer should update the state with the minimal amount of data to represent the situation.

assets/src/scripts/components/hydrograph/store.js
ayan-usgs: I'm wondering if it would make more sense to remove this in favor of a selector that figures out which markers to display from the `showSeries` and `legendMarkers` attributes.

assets/src/scripts/components/hydrograph/store.js
ayan-usgs: You know, I just made a very similar comment at around the same time this was posted. :ghost:

danielnaab: I just saw that 👍

assets/src/scripts/components/hydrograph/store.js
mbucknell: 👍

ayan-usgs: > Can legendMarkers also be a selector?

I was thinking about having the selector calculate the markers as part of figuring out which one's to display, so there wouldn't be a need for a selector specifically for the creation of the marker objects.

ayan-usgs:
@danielnaab -- Alright. I think this might be ready for another look.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -169,44 +172,52 @@ const plotMedianPoints = function (elem, {visible, xscale, yscale, medianStatsDa
...
+        .append('svg');
+
+
+    svg.call(link((elem, layout) => elem.attr('viewBox', `0 0 ${layout.width} ${layout.height}`), layoutSelector))
+        .call(link(addSVGAccessibility, createStructuredSelector({
+            title: state => state.title,
+            description: state => state.desc,
+            isInteractive: () => true
+        })))
+        .append('g')
```
danielnaab: Personal preference, I guess - but I like having each new node's attributes and children indented an extra level out, so you can visualize more quickly what element you're working on.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -169,44 +172,52 @@ const plotMedianPoints = function (elem, {visible, xscale, yscale, medianStatsDa
...
+            data: pointsSelector('current')
+        })))
+        .call(link(plotMedianPoints, createStructuredSelector({
+            visible: isVisibleSelector('medianStatistics'),
+            xscale: xScaleSelector('current'),
+            yscale: yScaleSelector,
+            medianStatsData: pointsSelector('medianStatistics')
+        })));
+
+    svg.call(link(plotLegend, createStructuredSelector({
```
danielnaab: Could this be chained like the other plot functions?

ayan-usgs: Technically no. It's as more that creating a legend on the svg before the lines were there objected with my graphing sensibilities. :alien: 

Irrespective of that, I can make that change.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: I'm cool with that.

assets/src/scripts/components/hydrograph/legend.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+        for (let seriesName of shownSeries) {
+            if (seriesName === 'compare' || seriesName === 'current') {
+                text = 'Current Year';
+                let domId = `ts-${seriesName}`;
+                let svgGroup = `${seriesName}-line-marker`;
+                if (seriesName === 'compare') {
+                    text = 'Last Year';
+                }
+                marker = defineLineMarker(domId, 'line', text, svgGroup);
```
danielnaab: I'm not sure, but I kind of feel like the actual DOM nodes should be created by the `drawSimpleLegend` function and the selectors just used for grabbing the state necessary to create the elements. But maybe that's not a distinction we should make? Selectors are pretty useful to use everywhere...

But to complete my thinking, what I mean is in `drawSimpleLegend` you could do something like:

```javascript
legendGroup.append(lineMarker('line', ....
// or maybe...
legendGroup.call(appendLineMarker, domId, 'line', text, svgGroup)
```

I'm curious your thoughts on that...

assets/src/scripts/components/hydrograph/store.js
```diff
@@ -94,6 +86,7 @@ export const timeSeriesReducer = function (state={}, action) {
         case 'ADD_TIMESERIES':
             // If data is valid
             if (action.data && action.data.values) {
+                let variableName = replaceHtmlEntities(action.data.variableName);
```
danielnaab: 👍

assets/src/scripts/components/hydrograph/store.spec.js
```diff
@@ -97,13 +98,17 @@ describe('Redux store', () => {
         it('should handle SET_MEDIAN_STATISTICS', () => {
             expect(timeSeriesReducer({}, {
                 type: 'SET_MEDIAN_STATISTICS',
-                medianStatistics: 'test data'
+                medianStatistics: {beginYear: '2000', endYear: '2010', values: ['a']}
```
danielnaab: Now that there's extra data on `medianStatistics`, I think it may be clearer on the action to flatten it like:

```
{
    type: 'SET_MEDIAN_STATISTICS',
    beginYear: '2000',
    endYear: '2010',
    values: ['a']
}
```

Although I've also seen this convention used in quite a few places, maybe we should do this everywhere...?

```
{
    type: 'SET_MEDIAN_STATISTICS',
    payload: {
        beginYear: '2000',
        endYear: '2010',
        values: ['a']
    }
}
```

Perhaps just ignore this for comment for now, and maybe we can discuss conventions like that after using this for awhile.

assets/src/scripts/components/hydrograph/legend.js
ayan-usgs: So, I was thinking about couple things when I designed that:

* I wanted `drawSimpleLegend` to have absolutely no what legend markers it had to work with. I was thinking that this would make it more general for when the legend gets more complicated. Though, the logic to create the markers has to live somewhere, so perhaps that's not a pertinent concern.
* My understanding of the a selector it's used to compute things from the state. So, I thought that since the DOM nodes that would appear in the legend are determined by the `showSeries` attribute in the state, it seemed to make sense to derive them in them in the selector.

assets/src/scripts/components/hydrograph/store.spec.js
danielnaab: Is it necessary to store begin and end years on the state, or could we derive that from the date ranges of the median state values?

ayan-usgs: The median values that we get make from the stats service is only a month and date without a year (though begin and end year are specified for each record in the returned RDB). I guess an alternative would be to include the begin and end year in with each of the median values... though having the begin and end year repeated 366 times in the values array seems a bit unnecessary.

assets/src/scripts/components/hydrograph/legend.js
danielnaab: Yes, it does work pretty well...

@mbucknell you have a thought? My gut is to keep all the dom creation in the functions you can trace directly off the main rendering function.

assets/src/scripts/components/hydrograph/store.spec.js
ayan-usgs: I end up having to put the current or previous year on the dates, so they will graph properly. So, if we were to use just the dates, we'd end up with things like `2017 - 2018`.

danielnaab: Oh, right...

danielnaab: Hmm... maybe median stats shouldn't be stored under tsData.

assets/src/scripts/components/hydrograph/legend.js
mbucknell: In the spirit of MVC, it seems like the selectors are part of the Model, whereas DOM is clearly part of the View. I haven't looked at this part of the code in detail but IMHO we should avoid DOM manipulation in the selector. Of course, I think this code merely creates the nodes and they are placed later. So I'm leaning more towards having the DOM nodes created in the rendering function rather than the selector.

ayan-usgs: Sure. Perhaps we can have the selector return an array of the things that need a legend. Then `drawSimpleLegend` function can generate the DOM nodes for those markers before appending them to svg legend group?

danielnaab: That seems good to me.

ayan-usgs: @danielnaab -- Reworked. A bit different than what I had described, but it think it's cleaner this way.

assets/src/scripts/components/hydrograph/store.spec.js
ayan-usgs: Per your other comment, I think we should think about how to structure the redux store.

PR #69, opened 2018-02-02
WDFN-78 - Redux integration

danielnaab:
I apologize for the size of this PR - it was more work than I expected.

The key ideas are:
- Selectors used to retrieve application state. Selectors are functions of the form `(state) => value`
- Computed selectors use the reselect library, which will cache selections as long as the dependencies don't change. Anything that's derived from state - such as the axes - should be accessed via a selector.
- Redux store is bound with the `provide` function to the D3 tree.
- View functions subscribe to state changes and redraw when dependent state changes, via the `link` function. Dependent state is passed to the link function as a selector.
- DOM events are routed to the Redux reducer via the `dispatch` function.

It's worth reading through the Redux docs: https://redux.js.org/
And also the reselect docs: https://github.com/reactjs/reselect

I also added awareness of Redux Dev Tools, which has Chrome and Firefox extensions you can install. It's really slick! https://github.com/zalmoxisus/redux-devtools-extension


assets/src/scripts/accessibility.js
```diff
@@ -8,7 +6,7 @@ const { select } = require('d3-selection');
  * @param {String} description
  * @param {Boolean} isInteractive
  */
-function addSVGAccessibility({svg, title, description, isInteractive}) {
+function addSVGAccessibility(svg, {title, description, isInteractive}) {
```
ayan-usgs: I'm still not sure when it's appropriate to include an argument in a destructuring target. What's the rationale for this change?

assets/src/scripts/components/hydrograph/points.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+
+/**
+ * Returns the current show state of a timeseries.
+ * @param  {Object}  state     Redux store
+ * @param  {String}  tsDataKey Timeseries key
+ * @return {Boolean}           Show state of the timeseries
+ */
+const isVisibleSelector = function (state, tsDataKey) {
```
ayan-usgs: Is there a reason anonymous functions are being used rather than named ones?

danielnaab: Only so they can be marked as `const`. I don't mind switching to named functions if you or Mary dislike it.

mbucknell: I'm not sure what this comment applies too. Are you talking about line 46?

ayan-usgs: Not really. I was more referring the this situation:
```
const myFunc = funct(x, y) {
     do some stuff;
}
```
versus
```
function myfunc(x, y) {
    do some stuff;
}

Comment (mbucknell) [reply assets/src/scripts/components/hydrograph/points.js]: So I really am fried. So the question is why isn't this written as function isVisibleSelector(state, tsDataKey)? I don't mind it being assigned to a const. In fact, that's the pattern I tried to use when doing standard JS except had to use var.

File: assets/src/scripts/accessibility.js (hunk 1 above)
Comment (danielnaab) [reply assets/src/scripts/accessibility.js]: When you use `d3.call`, it passes the current element as the first argument. Also, the `link` function will pass the additional arguments (derived from the selector it accepts) as the second argument... so the convention for our D3 "component" functions would be `(elem, options)`.

Comment (ayan-usgs) [reply assets/src/scripts/accessibility.js]: That makes a lot of sense.

File: assets/src/scripts/components/hydrograph/points.js (hunk 2 above)
Comment (ayan-usgs) [reply assets/src/scripts/components/hydrograph/points.js]: @mbucknell  -- That is pretty much the question.

I don't really mind the pattern in and of itself. It was more question aimed a probing the reason for using this pattern here, while the other modules seem to use `function someName()`.

Comment (mbucknell) [reply assets/src/scripts/components/hydrograph/points.js]: I think it also reinforces that functions are first class objects.

File: assets/src/scripts/lib/redux.js [hunk 3]
```diff
@@ -0,0 +1,86 @@
+// This code is derived from d3-redux: https://github.com/couchand/d3-redux
```
mbucknell: Why not just use npm install this?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -1,213 +1,130 @@
...
-        this._plotMedianPoints();
-
-        // Add the new time series
-        this._plotDataLine(this.plot, {xScale: xScale, yScale: this.scale.yScale}, 'compare');
+const drawMessage = function (elem, message) {
+    // Set up parent element and SVG
+    elem.innerHTML = '';
+    const alertBox = elem
+        .append('div')
+            .attr('class', 'usa-alert usa-alert-warning')
```
mbucknell: This really has nothing to do with this PR, but I think it would be more appropriate to use Information Status.

assets/src/scripts/accessibility.js
```diff
@@ -25,8 +23,10 @@ function addSVGAccessibility({svg, title, description, isInteractive}) {
  * @param {Array} columnNames - array of strings
  * @param {Array} data - array of array of strings
  */
-function addSROnlyTable({container, columnNames, data}) {
-    const table = select(container)
+function addSROnlyTable(container, {columnNames, data}) {
+    container.selectAll('table.usa-sr-only').remove();
```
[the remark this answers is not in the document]
danielnaab: If the link function calls this a second time, it would append a second table, so this removes the previous one (if it exists).

assets/src/scripts/components/hydrograph/axes.spec.js
```diff
@@ -37,19 +34,6 @@ describe('Chart axes', () => {
 
     it('axes appended', () => {
         // Should be translated
-        expect(svg.select('.x-axis').attr('transform')).toBe('translate(10, 10)');
-        expect(svg.select('.y-axis').attr('transform')).toBe('translate(0, 100)');
         expect(svg.select('.y-axis-label').text()).toEqual('Label title');
     });
-
-    it('tickValues should change with scale with new domain', () => {
```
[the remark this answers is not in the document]
danielnaab: I'm not sure how that would work, but can look into it.

assets/src/scripts/accessibility.js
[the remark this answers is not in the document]
mbucknell: We will have to be careful about this as we add more functionality. I noticed that you were using that pattern.

assets/src/scripts/lib/redux.js
danielnaab: I made modifications to it. Andrew and I discussed last week whether or not it's worth forking and maintaining a separate copy, but decided that given how simple it is, it makes more sense to just keep our own copy. I did offer suggestions for improvement to the original author, but he hasn't responded yet.

mbucknell: What kind of changes?

danielnaab: The author intended it to only be used via its fromState function with D3's data binding (data and datum functions). So to make it more general, I updated the connect function to pass the state, and added the link function to optimize the redraws.

PR #72, opened 2018-02-06
In median statistics parsing, convert values from string to float.

danielnaab:
I don't know if this was an issue before, but the extent calculation on the yAxis expects the value to be a number.


ayan-usgs:
> I don't know if this was an issue before, but the extent calculation on the yAxis expects the value to be a number.

I don't think that has caused me problems in the past. :thinking:

PR #73, opened 2018-02-06
Qualifiers WDFN-71

danielnaab:


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -39,20 +39,28 @@ const drawMessage = function (elem, message) {
...
             .classed('line', true)
-            .attr('id', elemId)
-            .attr('d', line().x(d => xScale(d.time))
-                             .y(d => yScale(d.value)));
+            .classed('approved', line.classes.approved)
+            .classed('estimated', line.classes.estimated)
+            .attr('data-title', tsDataKey)
+            .attr('id', `ts-${tsDataKey}`)
+            .attr('d', tsLine);
+    }
```
danielnaab: Each line is styled with CSS classes.

assets/src/scripts/components/hydrograph/points.js
```diff
@@ -27,15 +27,58 @@ const isVisibleSelector = memoize(tsDataKey => (state) => {
...
+
+            // Add this point to the current line.
+            lines[lines.length - 1].points.push(pt);
+
+            // Cache the classes for the next loop iteration.
+            lastClasses = lineClasses;
+        }
+
+        return lines;
+    }
```
danielnaab: This will split on estimated/approved status, and ignore null values (masked data). For the work to draw the masks, this selector will need to be updated to include those masked points in the response.

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -72,20 +72,6 @@ function createYScale(tsData, showSeries, ySize) {
...
- * @param  {Number} ySize Y range of scale
- * @return {Object}        {xScale, yScale}
- */
-function createScales(tsData, showSeries, xSize, ySize) {
-    const xScale = createXScale(tsData.current, xSize);
-    const yScale = createYScale(tsData, showSeries, ySize);
-
-    return {xScale, yScale};
-}
-
```
danielnaab: This was only used in the tests.

assets/src/scripts/models.js
```diff
@@ -48,19 +48,23 @@ export function getTimeseries({sites, params=['00060'], startDate=null, endDate=
...
                                 return {
                                     time: date,
-                                    value: parseFloat(value.value),
-                                    label: `${formatTime(
-                                        date)}\n${value.value} ${series.variable.unit.unitCode}`
+                                    value: value !== noDataValue ? parseFloat(datum.value) : null,
+                                    qualifiers: datum.qualifiers,
+                                    approved: datum.qualifiers.indexOf('A') > -1,
+                                    estimated: datum.qualifiers.indexOf('E') > -1,
+                                    label: `${formatTime(date)}\n${datum.value} ${series.variable.unit.unitCode} (Qualifiers: ${datum.qualifiers.join(', ')})`
```
danielnaab: As I understand from a conversation with Brad, "approved" has the "A" qualifier, and pending has "P". Rather than test for "P", this just checks for A, as one or the other should always be set.

assets/src/scripts/models.js
```diff
@@ -150,13 +155,10 @@ export function parseMedianData(medianData, timeSeriesStartDateTime, timeSeriesE
...
-                else {
+                if (!(month == 1 && day == 29)) {
                     data.push(median);
                 }
-            }
-            else {
+            } else {
                 data.push(median);
             }
         }
```
danielnaab: Just clearing some eslint messages here.

assets/src/scripts/components/hydrograph/points.js
ayan-usgs: Sounds good.

PR #75, opened 2018-02-07
Wdfn 61 - rescale fonts for mobile

mbucknell:
Added a new property width to the store along with the new action, RESIZE_TIMESERIES_PLOT. These are used to redraw the timeseries plot whenever the window is resized.


assets/src/scripts/components/hydrograph/index.spec.js
```diff
@@ -237,111 +242,6 @@ describe('Hydrograph charting module', () => {
     });
 });
 
-
```
mbucknell: Removed because these constants are no longer used.

assets/src/styles/components/_hydrograph.scss
```diff
@@ -50,16 +50,24 @@
...
+            .tick {
+                line {
+                    fill: none;
+                    stroke: lightgrey;
+                    stroke-opacity: 0.7;
+                }
             }
-            text {
+        }
+        .x-axis {
```
mbucknell: I found the x tick marks to be nearly invisible with using the lightgrey color. Made these black so they are easier to see.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -231,13 +229,22 @@ const attachToNode = function (node, {siteno} = {}) {
...
         .call(provide(store))
-        .call(timeSeriesGraph)
         .select('.hydrograph-last-year-input')
             .on('change', dispatch(function () {
                 return Actions.toggleTimeseries('compare', this.checked);
             }));
+    plotSelect.append('div')
+            .attr('class', 'hydrograph-container')
+            .style('padding-bottom', ASPECT_RATIO_PERCENT)
+            .call(link(timeSeriesGraph, layoutSelector));
```
danielnaab: Nesting calls to `link()` will be problematic, because the previously rendered nodes will still have their store subscriptions active.

We might eventually need a more general way to bind attribute values to nodes, but I think you can achieve the same thing with a call(link()) up on the svg node, like:

```
elem.append('svg')
    .call(link((elem, layout) => elem.attr('viewBox', `0 0 ${layout.width} ${layout.height}`), layoutSelector))
```

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -231,13 +229,22 @@ const attachToNode = function (node, {siteno} = {}) {
...
-    select(node)
+    store.dispatch(Actions.resizeTimeseriesPlot(node.offsetWidth));
+    let plotSelect = select(node);
+    plotSelect
         .call(provide(store))
-        .call(timeSeriesGraph)
         .select('.hydrograph-last-year-input')
             .on('change', dispatch(function () {
                 return Actions.toggleTimeseries('compare', this.checked);
             }));
```
danielnaab: Separate issue, but I think we should consider rendering the checkbox here rather than in the Jinja2 template, and tie its checked state to something in the Redux store.

mbucknell: At the moment, I would rather not render it here, but we can easily add code here to set it's state base on the Redux store. This page doesn't have a lot of tools yet. In the future, I can see developing some "components" but I would like to wait on that until other things settle down.

danielnaab: I'm just imaging an `.append('input')`, not a complex component.

assets/src/scripts/components/hydrograph/layout.js
```diff
@@ -1,11 +1,30 @@
 // Define width, height and margin for the SVG.
 // Use a fixed size, and scale to device width using CSS.
```
danielnaab: Suppose the comment should be updated

mbucknell: 👍

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -168,13 +168,10 @@ const plotMedianPoints = function (elem, {visible, xscale, yscale, medianStatsDa
...
-    elem.append('div')
-        .attr('class', 'hydrograph-container')
-        .style('padding-bottom', ASPECT_RATIO_PERCENT)
-        .append('svg')
-            .attr('preserveAspectRatio', 'xMinYMin meet')
-            .attr('viewBox', `0 0 ${WIDTH} ${HEIGHT}`)
+const timeSeriesGraph = function (elem, layout) {
+    elem.selectAll('svg').remove();
+    elem.append('svg')
+            .attr('viewBox', `0 0 ${layout.width} ${layout.height}`)
```
danielnaab: Was the extra indent level here intentional?

assets/src/scripts/components/hydrograph/index.js
mbucknell: For usability, you need to make sure the label has the for attribute set, etc. I agree that we could do it and we may want to do it in the future but I don't see a driving reason to change it

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -95,8 +95,8 @@ const plotTooltips = function (elem, {xScale, yScale, data}) {
 
     elem.append('rect')
         .attr('class', 'overlay')
-        .attr('width', WIDTH)
-        .attr('height', HEIGHT)
+        .attr('width', layout.width)
+        .attr('height', layout.height)
```
danielnaab: I wonder if this could be set to 100% somehow, to avoid the redraw on change?

mbucknell: I'll give a try.

mbucknell:
@danielnaab , I am trying to understand better how the link command works and see your point about re-renderer the lines since that is definitely happening right now. I

assets/src/scripts/components/hydrograph/index.js
mbucknell: I don't understand this: " the previously rendered nodes will still have their store subscriptions active". Can you explain? Does the state get messed up?

danielnaab: The state will be okay, there will just be unneeded renders over the old, deleted DOM nodes.

The `connect` function registers a listener to Redux state changes and calls the provided function on state change. The `link` function wraps the connect function, to skip the call to the function if the selector it accepts returns the same object as last time.

When you do `.call(connect(..))` or `.call(link())`, in both cases the original D3 selector provided by `call` is saved, and passed in when the calls to actual rendering function is made. So what ends up happening, is those functions are called with the old D3 DOM selections, even though they're not in the document any more.

These higher order functions are hard to keep straight... I can clear up more on Slack if that doesn't make sense.

mbucknell: I put in lots of breakpoints in and can see the extra render calls with the way I was doing it. Changing to just using the link on the code to add the viewBox makes way more sense. Thanks for explaining.

mbucknell: 💡 - I finally get it!

ayan-usgs: > So what ends up happening, is those functions are called with the old D3 DOM selections, even though they're not in the document any more.

I think that mostly makes sense, though I'm confused by the above statement.... :thinking:

mbucknell:
Just a clarification: The time series lines do get rerendered because the xscale and yscale are upddated. But it's all done through  Redux state management.

PR #82, opened 2018-02-12
fix incorrect legend positioning on window resize

ayan-usgs:


assets/src/scripts/components/hydrograph/legend.js
```diff
@@ -69,7 +71,16 @@ function drawSimpleLegend(svg,
     }
     // center the legend group in the svg
     let legendBBox = legend.node().getBBox();
-    let legendXPosition = (svgBBox.width - legendBBox.width) / 2;
+
+    let legendXPosition;
```
mbucknell: nit: Could simplify this with tertiary operator: 
```
const svgWidth width ? width : svgBBox.width;
const legendXPosition = (svgWidth  - legendBBox.width) / 2;

PR #87, opened 2018-02-13
Wdfn 80 - Show/Hide Median details on click on median circles

mbucknell:


danielnaab:
It would be nice if the median stat circle in the legend was also clickable (or remove the hand cursor from it).

mbucknell:
Didn't notice it was clickable. Let me see how easy it is to make it really clickable. I like that idea!

PR #88, opened 2018-02-13
Changed to use title and desc tags rather than attributes and added c…

mbucknell:
…ode to remove those attributes before building the new DOM.

### DIFF
--- a/assets/src/scripts/accessibility.js
+++ b/assets/src/scripts/accessibility.js
@@ -7,8 +7,7 @@
  * @param {Boolean} isInteractive
  */
 function addSVGAccessibility(svg, {title, description, isInteractive}) {
-    svg.select('title').remove();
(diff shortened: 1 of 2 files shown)


assets/src/scripts/accessibility.js
```diff
@@ -7,9 +7,14 @@
  * @param {Boolean} isInteractive
  */
 function addSVGAccessibility(svg, {title, description, isInteractive}) {
-    svg.attr('title', title)
-        .attr('desc', description)
-        .attr('aria-labelledby', 'title desc');
+    svg.select('title').remove();
+    svg.select('desc').remove();
```
danielnaab: fyi, if you haven't seen it, selections support commas: `svg.select('title, desc').remove()`

mbucknell: 👍

PR #89, opened 2018-02-13
WDFN-67 - Create a symlog-like graph, using a "compound scale".

danielnaab:
This introduced an issue with tooltips on sites that have 100% masked data. @mbucknell, since you've been working in there, I'll wait for the merge to see if the issue needs resolution.

### DIFF
--- a/assets/src/scripts/components/hydrograph/scales.js
+++ b/assets/src/scripts/components/hydrograph/scales.js
@@ -1,7 +1,8 @@
 const { extent } = require('d3-array');
-const { scaleLinear, scaleTime } = require('d3-scale');
+const { scaleTime } = require('d3-scale');
 const { createSelector, defaultMemoize: memoize } = require('reselect');
 
+const { default: scaleSymlog } = require('../../lib/symlog');
 const { layoutSelector, MARGIN } = require('./layout');
 
 const paddingRatio = 0.2;
@@ -14,8 +15,19 @@ const paddingRatio = 0.2;
  */
 function extendDomain(domainExtent) {
     const padding = paddingRatio * (domainExtent[1] - domainExtent[0]);
+
+    // If all values are above zero, make zero the lower bound
+    if (domainExtent[0] >= 0 && domainExtent[1] >= 0) {
+        return [
+            Math.max(0, domainExtent[0] - padding),
+            domainExtent[1] + padding
+        ];
+    }
+
+    // If we have values less than zero, keep 20% padding around both sides.
     return [domainExtent[0] - padding, domainExtent[1] + padding];
 }
+
 
 /**
  * Create an x-scale oriented on the left
@@ -69,10 +81,9 @@ function createYScale(tsData, showSeries, ySize) {
         yExtent = [0, 1];
     }
 
-    // yScale is oriented on the bottom
-    return scaleLinear()
-        .range([ySize, 0])
-        .domain(yExtent);
+    return scaleSymlog()
+        .domain(yExtent)
+        .range([ySize, 0]);
 }
 
 
--- a/assets/src/scripts/components/hydrograph/scales.js
+++ b/assets/src/scripts/components/hydrograph/scales.js
@@ -9,23 +9,19 @@ const paddingRatio = 0.2;
 
 
 /**
- *  Return domainExtent padded on both ends by paddingRatio
- *  @param {Array} domainExtent - array of two numbers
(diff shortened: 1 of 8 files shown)


danielnaab:
There is a corresponding d3-scale PR here: https://github.com/d3/d3-scale/pull/134

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -14,9 +15,20 @@ const paddingRatio = 0.2;
  */
 function extendDomain(domainExtent) {
     const padding = paddingRatio * (domainExtent[1] - domainExtent[0]);
+
+    // If all values are above zero, make zero the lower bound
+    if (domainExtent[0] >= 0 && domainExtent[1] >= 0) {
+        return [
+            Math.max(0, domainExtent[0] - padding),
```
mbucknell: Do we have a preference for using Math.max over d3.max? I have been using d3.max in the tooltip code for basically the same purpose as here where I'm finding the max of two numbers. I used d3.max because I was already using it in the package to find the maximum of an array.

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -14,9 +15,20 @@ const paddingRatio = 0.2;
...
+            domainExtent[1] + padding
+        ];
+    }
+
+    // If we have values less than zero, keep 20% padding around both sides.
```
mbucknell: Style thing, but I think having one return statement is cleaner.

assets/src/scripts/components/hydrograph/scales.js
ayan-usgs: I personally don't have a preference, though I've been using `Math` methods recently (e.g. `Math.abs`), but that's because I couldn't find a d3 absolute value method.

assets/src/scripts/lib/compound.js
```diff
@@ -0,0 +1,47 @@
+export default function compound() {
+  var scales = [].slice.call(arguments);
```
ayan-usgs: I'm just curious about the `var` vs `let` here?

Also, it seems that this stylistically a bit different from the rest of the javascript... :thinking:

danielnaab: This is a copy-paste of the code I submitted to d3-scale, so I wrote it in the D3 style.

ayan-usgs: Ah. Makes sense.

assets/src/scripts/components/hydrograph/scales.js
danielnaab: I wasn't aware of d3.max - curious if it does something different than the standard one, which I would assume would be pretty optimized.

assets/src/scripts/components/hydrograph/scales.js
danielnaab: I like that when there's cleanup to do before returning, but otherwise I tend to prefer the "return early" notion of weeding out the edge cases first.

mbucknell: In this case though there really are only have two paths, in which case I think it's clearer to have an else clause.

assets/src/scripts/components/hydrograph/scales.js
danielnaab: In `d3-scale, only `Math.max` is used... From the docs, I assume there's specific cases where `d3.max` is useful but I don't think we need it here.

assets/src/scripts/lib/compound.js
```diff
@@ -0,0 +1,47 @@
...
+    if (values[0] > values[1]) range = range.slice().reverse();
+    return range;
+  };
+
+  scale.copy = function() {
+    return compound.apply(null, scales.map(function(s) { return s.copy(); }));
+  };
+
+  scale.scales = function(_) {
+    return arguments.length ? (scales = _, scale) : scales;
```
mbucknell: I'm not understanding this syntax. Can you describe what this does?

danielnaab: If the function is called without arguments, it returns the scales array. Otherwise, it sets the array to the passed in parameters and then returns the scale object, enabling function chaining. That clause is using the comma operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator

mbucknell: Thanks. Really need to read up on ES6 syntax.

PR #92, opened 2018-02-15
Use full qualifier description and refactored tooltips so text boxes can be placed anywhere in the svg

mbucknell:
The refactoring also allowed more unit tests to be written since event handlers now only update state.

### DIFF
--- a/assets/src/scripts/components/hydrograph/tooltip.js
+++ b/assets/src/scripts/components/hydrograph/tooltip.js
@@ -2,6 +2,10 @@
 const { max, bisector } = require('d3-array');
 const { mouse } = require('d3-selection');
 
+const { dispatch } = require('../../lib/redux');
+
+const { Actions } = require('./store');
+
 const maxValue = function (data) {
     return max(data.map((datum) => datum.value));
 };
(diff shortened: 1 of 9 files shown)


assets/src/scripts/components/hydrograph/tooltip.js
```diff
@@ -107,63 +173,41 @@ const createTooltip = function(elem, {xScale, yScale, compareXScale, currentTsDa
...
-                    yScale: yScale,
-                    tsDatum: compareData.datum
-                });
-            }
-        });
+        .on('mouseover', dispatch(function() {
+            return Actions.setTooltipTime(
+                xScale.invert(mouse(elem.node())[0]),
+                isCompareVisible ? compareXScale.invert(mouse(elem.node())[0]) : null
+            );
```
danielnaab: It's too bad the scales need to be different for each line, but given that I'm thinking doing the xScale.invert() in the code that moves the tooltip circle would be a (tiny) bit simpler, so only the single mouse location needs to be stored in state.

mbucknell: It's true that I could put the mouse location, but given that the tooltip text is in a different group then the tooltip focus, I think it's cleare to store the actual time.

PR #93, opened 2018-02-15
Wdfn 62 masking qualifiers

ayan-usgs:


assets/src/scripts/components/hydrograph/layout.js
```diff
@@ -7,7 +7,7 @@ const ASPECT_RATIO_PERCENT = `${100 * ASPECT_RATIO}%`;
 const MARGIN = {
     top: 20,
     right: 75,
-    bottom: 50,
+    bottom: 70,
```
ayan-usgs: There will be be some more fanciness around for the legend to be responsive, but there's already a separate task for that ironing out issues on the legend.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -49,21 +51,93 @@ const plotDataLine = function (elem, {visible, lines, tsDataKey, xScale, yScale}
...
+                .datum(line.points)
+                .classed('line', true)
+                .classed('approved', line.classes.approved)
+                .classed('estimated', line.classes.estimated)
+                .attr('data-title', tsDataKey)
+                .attr('id', `ts-${tsDataKey}`)
+                .attr('d', tsLine);
+        }
+        else {
+            let maskCode = line.classes.dataMask.toLowerCase();
```
mbucknell: We seem to be using let in places where we should use const. Except for maskGroup, I think these all should be const.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -49,21 +51,93 @@ const plotDataLine = function (elem, {visible, lines, tsDataKey, xScale, yScale}
...
+            let maskDisplayName = MASK_DESC[maskCode].replace(' ', '-').toLowerCase();
+            let xMaskExtent = extent(line.points, d => d.time);
```
mbucknell: Nit: Rather than declare an intermediate variable, maybe combine lines 67 and 68?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -49,21 +51,93 @@ const plotDataLine = function (elem, {visible, lines, tsDataKey, xScale, yScale}
...
+                    .attr('class', `${tsDataKey}-mask-group`);
+
+            maskGroup.append('rect')
+                .attr('x', xScale(xDomainStart))
+                .attr('y', yScale(yRangeEnd))
+                .attr('width', xScale(xDomainEnd) - xScale(xDomainStart))
+                .attr('height', Math.abs(yScale(yRangeEnd)- yScale(yRangeStart)))
+                .attr('class', `mask ${maskDisplayName}-mask`);
+
+            let patternId = tsDataKey === 'compare' ? 'url(#hash-135)' : 'url(#hash-45)';
```
mbucknell: I wonder if the hashes should be constants at the module level so that they aren't buried in the code?

assets/src/scripts/components/hydrograph/markers.js
```diff
@@ -13,11 +13,45 @@ function circleMarker({r, x, y, domId=null, domClass=null}) {
     if (domClass !== null) {
         circle.attr('class', domClass);
     }
-    return circle;
-}
+    if (fill != null) {
```
mbucknell: Why are you drawing a second circle instead of adding the fill to the first circle?

assets/src/scripts/components/hydrograph/markers.js
```diff
@@ -13,11 +13,45 @@ function circleMarker({r, x, y, domId=null, domClass=null}) {
...
+        .attr('width', width)
+        .attr('height', height);
+    if (domId !== null) {
+        rectangle.attr('id', domId);
+    }
+    if (domClass !== null) {
+        rectangle.attr('class', domClass);
+    }
+
+    if (fill !== null) {
```
mbucknell: Same question as above?

assets/src/scripts/components/hydrograph/points.js
```diff
@@ -37,25 +54,37 @@ const lineSegmentsSelector = memoize(tsDataKey => createSelector(
...
             };
-
+            if (pt.value === null) {
+                let qualifiers = new Set(pt.qualifiers.map(q => q.toLowerCase()));
+                // current business rules specify that a particular data point
+                // will only have at most one masking qualifier
+                let maskIntersection = new Set([...masks].filter(x => qualifiers.has(x)));
+                lineClasses.dataMask = [...maskIntersection][0]
+            }
+            else if (pt.value !== null) {
```
mbucknell: This can just be an else {}

mbucknell:
For your consideration: The legend module functions, drawSimpleLegend and createLegendMarkers  might benefit from breaking down into smaller functions (could just be local to the module). Not something that needs to be done know for sure.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: Okay.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: Yep.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: I think that would be good.

assets/src/scripts/components/hydrograph/markers.js
ayan-usgs: So, there might be some SVG shape capability, that I haven't discovered yet, but while `fill` works pretty good and much as you would expect if it was a color, I couldn't figure out a way to have a pattern that has a background color.

What I ended up doing was creating shape with the color I want, and then overlayed another shape over it that has the pattern. This approach what's creates the colored mask blocks with the diagonal hashing in both the plot and legend.

assets/src/scripts/components/hydrograph/points.js
ayan-usgs: Yep.

ayan-usgs:
> For your consideration: The legend module functions, drawSimpleLegend and createLegendMarkers might benefit from breaking down into smaller functions (could just be local to the module). Not something that needs to be done know for sure.

I think that would be a good idea. I'll write up a task for it.

assets/src/scripts/components/hydrograph/points.js
ayan-usgs: Now that I look at it more closely, I think that's a merge error...

assets/src/scripts/components/hydrograph/markers.js
mbucknell: I think this is worthy of a comment in the code

PR #95, opened 2018-02-16
include site visits and parameter code

ayan-usgs:


waterdata/templates/monitoring_location.html
```diff
@@ -78,8 +79,9 @@ <h1>Station Name: {{ stations[0].station_nm }}</h1>
                                     <tbody>
                                         {% for series in site_dataseries %}
                                             <tr>
-                                                <th scope="row">{{ series.parm_cd.name }}</td>
+                                                <td scope="row">{{ series.parm_cd.name if series.parm_cd.name else '<sub>n/a</sub>' | safe }}</td>
```
danielnaab: There might be some scenario where we want the name escaped here (and below)... I think we should error on the side of caution and use `{% if series.parm_cd.name %}series.parm_cd.name{% else %}<sub>n/a</sub>{% endif %}` instead.

ayan-usgs: Adjusted.

PR #97, opened 2018-02-20
Header fix and eliminate duplicate ids in DOM

mbucknell:
And a few  style cleanups


mbucknell:
I'm puzzled that Codacy is not picking this up. It seems like it either analyzes the javascript or it analyzes the python. It looks like the SaltPyLint plugin is now working because it's finding the issue I inserted.

PR #98, opened 2018-02-20
WDFN-74 Show all IV parameter codes

danielnaab:


.eslintrc
```diff
@@ -18,6 +18,7 @@
   "rules": {
     "comma-dangle": [2,"never"],
     "linebreak-style": ["error", "unix"],
+    "no-console": "off",
```
mbucknell: Do you think it would be worthwhile to bring in a logging library? I have used loglevel.js, https://www.npmjs.com/package/loglevel, in the past.

danielnaab: Yes that looks like a good library. In this case, I was envisioning using it just to notify us during development of errors.... I had an exception that was being silenced by a promise and needed to see its contents. But seems we should consider a comprehensive approach.

assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -0,0 +1,86 @@
...
+                .text(d => d);
+
+    table.append('tbody')
+        .selectAll('tr')
+        .data(availableTimeseries)
+        .enter().append('tr')
+            .classed('selected', parm => parm[1].selected)
+            .on('click', dispatch(function (parm) {
+                if (!parm[1].selected) {
+                    return Actions.setCurrentParameterCode(parm[0]);
```
ayan-usgs: Could you elaborate on what is in the `param` array?

danielnaab: Yes - `availableTimeseries` is an array of `(parameter code, metadata)` pairs... where `metadata` is an object with all the necessary info needed by this function.

PR #99, opened 2018-02-21
replace UTC offset with timezone

ayan-usgs:
I think this is all that needs to happen. The times that come back from the service are already converted to the browser's local time.


assets/src/scripts/models.js
```diff
@@ -75,7 +76,7 @@ export function getTimeseries({sites, params=null, startDate=null, endDate=null}
                             qualifiers: datum.qualifiers,
                             approved: datum.qualifiers.indexOf('A') > -1,
                             estimated: datum.qualifiers.indexOf('E') > -1,
-                            label: `${formatTime(date)}\n${value || ''} ${value ? series.variable.unit.unitCode : ''} (${qualifierDescriptions.join(', ')})`
+                            label: `${formatTime(date)} ${tzAbbrev}\n${value || ''} ${value ? series.variable.unit.unitCode : ''} (${qualifierDescriptions.join(', ')})`
```
mbucknell: I wonder how hard it would be to add some tests for the label? Not suggesting you do this (I didn't do it for the qualifierMapping), but if you have some time this might be a good thing to add.

ayan-usgs: I extended on of the tests to check the label.

PR #100, opened 2018-02-21
Wdfn 52

mbucknell:
Added google analytics to the project. The tracking code can be specified with GA_TRACKING_CODE. We can also control whether the USGS analytics are included with ENABLE_USGS_GA.

Also added the autotrack.js library, see https://github.com/googleanalytics/autotrack/. Adds the ability to track various events automatically if configured to do so. I decided to import the entire autotrack.js library (8K) rather than build a custom one. As we go on, I'm thinking that different pages will be using different autotrack capabilities. I have added just two, outboundLinkTracker and eventTracker. I have added the markup to track events for the compare toggle and the time series selection list.

In addition to this PR, I will also be updating the Ansible Playbook to set the configuration parameters appropriately for different tiers.


danielnaab:
Cool library

PR #104, opened 2018-02-22
Wdfn 104

mbucknell:


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -246,30 +246,29 @@ const timeSeriesGraph = function (elem) {
...
+            describeById: () => 'current-time-series-sr-desc',
             describeByText: () => 'current time series data in tabular format'
     })));
-
     elem.append('div')
         .call(link(addSROnlyTable, createStructuredSelector({
             columnNames: createSelector(
                 titleSelector,
-                (title) => [`Median ${title}`, 'Time']
+                (title) => [title, 'Time', 'Qualifiers']
```
ayan-usgs: Not something that needs to be done now, but I wonder if we need to figure out some way to explain the qualifiers in an accessible way. A `P` in the table doesn't really say much about why it's there or what it means.

mbucknell: The same thought occurred to me. Maybe it would be best to but the qualifier description directly in the time series data and then extract it. I think I'll go ahead and do that.

danielnaab: @mbucknell you can skip that - with the normalizr state refactoring, I added the qualifier description into the state.

mbucknell: Just did it but I'll merge this as is so you don't have merge hell.

PR #105, opened 2018-02-22
Wdfn 93 monitoring location questions link

ayan-usgs:


waterdata/commands/lookup_generation/__init__.py
```diff
@@ -89,6 +88,66 @@
               'aw': {'name': 'USGS Active Groundwater Level Network Site'},
               'ad': {'name': 'USGS Annual Water Data Reports Site'}}
 
+US_STATES = [{'name': 'Alabama', 'abbreviation': 'AL'},
```
mbucknell: Rather the generate a separate lookup file, should we just put this in constants.py?

ayan-usgs: I just moved it there this morning. :stuck_out_tongue_closed_eyes:

PR #108, opened 2018-02-27
Wdfn 97 - Make Legend More responsive

mbucknell:
The legend has been reworked to show fewer markers and to have each time series on its own line. This removes the reflowing of the legend when clicking the compare toggle and makes layout simpler.


assets/src/scripts/components/hydrograph/legend.js
```diff
@@ -1,188 +1,164 @@
...
 
 
 /**
- * Create a simple horizontal legend
+ * Create a simple legend
  *
- * @param svg
- * @param legendMarkers
- * @param width
- * @param startingXPosition
```
mbucknell: I removed all of the positioning parameters. The layout is pretty specific to what we are doing here. I tried playing around with values with not so nice results. Put the ones we still needed as function const.

assets/src/scripts/components/hydrograph/timeseries.js
```diff
@@ -7,7 +7,7 @@ const { createSelector } = require('reselect');
 const formatTime = timeFormat('%c %Z');
 
 const MASK_DESC = {
-    ice: 'Ice',
+    ice: 'Ice Affected',
```
mbucknell: Jim requested this be changed to Ice Affected

assets/src/scripts/components/hydrograph/legend.js
ayan-usgs: Sounds good.

PR #110, opened 2018-02-28
fix typo and harden code if a district code is not available

ayan-usgs:


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,9 @@
+{
```
mbucknell: This file should not be part of the git repo

ayan-usgs: As yes... the perils of using `git add --all`.

PR #115, opened 2018-03-07
WDFN-103 Multiple time-series per parameter code

danielnaab:


assets/src/scripts/components/hydrograph/store.js
```diff
@@ -1,43 +1,57 @@
+const merge = require('lodash/merge');
```
mbucknell: I don't think this got added to package.json.

danielnaab: Oh, must be a dependency of another package we use. Will add.

assets/src/scripts/components/hydrograph/schema.js
```diff
@@ -0,0 +1,130 @@
...
+            variableName: replaceHtmlEntities(variable.variableName),
+            variableCode: variable.variableCode[0]
+        };
+    }
+});
+
+// timeSeries schema
+const qualifier = new schema.Entity('qualifiers', {}, {idAttribute: 'qualifierCode'});
+const method = new schema.Entity('methods', {}, {idAttribute: 'methodID'});
+const timeSeries = tsKey => new schema.Entity('timeSeries', {
```
mbucknell: Style issue: I don't think I like using arrow functions when defining a function object. Is there a reason you do so?

danielnaab: Just conciseness. But looking now, those should all perhaps be memoized..

mbucknell: That's a great idea!

assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -129,12 +131,21 @@ export const plotSeriesSelectTable = function (elem, {availableTimeseries, layou
                 // under the appropriate column headers
                 if (!screenSizeCheck) {
                     tr.append('td')
-                        .html(parm => parm[1].currentYear ? '<i class="fa fa-check" aria-label="Current year data available"></i>' : '');
+                        .html(parm => {
+                            const subScript = parm[1].currentYear > 1 ? `<sub>${parm[1].currentYear}</sub>` : '';
```
mbucknell: Might be more concise if we have a local function which creates the html. We also lose the subscripts in the tooltip

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -129,32 +142,37 @@ const plotSvgDefs = function(elem) {
...
 
 
-const plotLegend = function(elem, {displayItems, layout}) {
-    elem.select('.legend').remove();
-    let plotMarkers = createLegendMarkers(displayItems);
-    drawSimpleLegend(elem, plotMarkers, layout);
+const timeSeriesLegend = function(elem) {
+    elem.append('div')
+        .attr('class', 'hydrograph-container')
+        .append('svg')
```
mbucknell: Why change to drawing the legend in a separate SVG?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -294,4 +363,4 @@ const attachToNode = function (node, {siteno} = {}) {
 };
 
 
-module.exports = {attachToNode, timeSeriesGraph};
+module.exports = {attachToNode, timeSeriesLegend, timeSeriesGraph};
```
mbucknell: I'm not sure I like that we are exporting timeSeriesLegend and timeSeriesGraph only for testing. What do others think? I think they can be tested particularly now that we have separated the state from the rendering without exporting these functions.

assets/src/scripts/components/hydrograph/index.js
mbucknell: We will need to add at least some of the accessibility features to it such as a title and description.

assets/src/scripts/components/hydrograph/parameters.js
mbucknell: That's not what I meant. We could have a local function that creates the checkmark with the appropriate subscript.

assets/src/scripts/components/hydrograph/index.js
danielnaab: Yes, missed that.

assets/src/scripts/components/hydrograph/parameters.js
danielnaab: I caught that - but I was wondering what you meant by the reference to tooltip

mbucknell: Must have missed it. The subscripts are on the small screen tooltip.

danielnaab: Ah, I see. Yep, one more thing..

PR #117, opened 2018-03-08
Wdfn 104 fix to incorrect last year marks

ayan-usgs:


danielnaab:
I wonder if filtering out the empty time series in the selector would be a better approach? It would avoid a circumstance where current, compare, and median series have no points but the row still shows up in the table.

danielnaab:
As far as I'm aware now, we shouldn't do anything with empty time series anywhere in the graph, so perhaps the `allTimeSeriesSelector` could be updated to filter them out on the top-level, and use that in the `availableTimeseriesSelector`.

ayan-usgs:
> As far as I'm aware now, we shouldn't do anything with empty time series anywhere in the graph, so perhaps the allTimeSeriesSelector could be updated to filter them out on the top-level, and use that in the availableTimeseriesSelector.

That's a good idea.

assets/src/scripts/components/hydrograph/timeseries.js
```diff
@@ -82,7 +82,22 @@ export const methodsSelector = state => state.series.methods;
...
+            for (let key of Object.keys(stateTimeSeries)) {
+                const ts = stateTimeSeries[key];
+                if (ts.hasOwnProperty('points') && ts.points.length > 0) {
+                    timeSeries[key] = ts;
+                }
+            }
+        }
+        return timeSeries;
+    }
+);
```
danielnaab: I'm running this locally and found an issue on this site: http://localhost:5050/monitoring-location/251549080251200

The problem seems to be that timeSeriesSelector assumes that all the series in the "collection" are returned from allTimeSeriesSelector.

Yesterday we were talking about how the collection and request selectors are confused. Maybe you could just get rid of them as a part of this? The `state.series.timeSeries` elements have a `tsKey` attribute that can be used to filter by current, compare, etc.

ayan-usgs: @danielnaab: I might merge these changes in and remove those to selectors in a separate PR.

danielnaab: 👍

PR #118, opened 2018-03-08
WDFN-117 - Make y-axis ticks and label prettier

danielnaab:


assets/src/scripts/utils.js
```diff
@@ -69,6 +69,55 @@ export function deltaDays(date1, date2) {
...
-    let itemsEqual = [...set1].every(x => {return set2.has(x)});
+    let itemsEqual = [...set1].every(x => {
+        return set2.has(x);
+    });
     return sizeEqual && itemsEqual;
 }
+
+/**
+ * Wrap long svg text labels into multiple lines.
+ * From: https://bl.ocks.org/ericsoco/647db6ebadd4f4756cae
```
danielnaab: Code courtesy bl.ocks.org, so I didn't write any tests. Hopefully that's not a bad excuse?

mbucknell: It's a bad excuse 😁 .  Should probably add some and make it more ES6 friendly.

ayan-usgs: I concur with that sentiment.

assets/src/scripts/utils.js
```diff
@@ -69,6 +69,74 @@ export function deltaDays(date1, date2) {
...
+        let line = [];
+        let lineCount = 0;
+        for (const word of textContent.split(/\s+/)) {
+            // Add this word to the line
+            line.push(word);
+            tspan.text(line.join(' '));
+
+            // If we exceeded the line width, remove the last word from the array
+            // and append this tspan to the DOM node.
+            if (tspan.node().getComputedTextLength() > width) {
```
ayan-usgs: This is a pretty cool way of doing of doing the text wrapping.

PR #119, opened 2018-03-08
Wdfn 105 - Connect to FIM

mbucknell:
This looks larger than it is. Moved store.js to the top level scripts directory along with schema.js. There are two new modules: floodSlider.js and floodData.js and map.js has changes to add the FIM layers.

Also added postccs using auto-prefixer to the build step for css as recommended by uswds folks.


assets/src/scripts/components/floodSlider.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const floodSlider = function(node) {
+    const SLIDER_ID = 'fim-slider';
+    let sliderContainer = node.append('div')
+        .attr('class', 'slider-wrapper')
+        .property('hidden', true);
+    sliderContainer.append('label')
+        .attr('for', SLIDER_ID)
+        .html(`Gage Height: <output class="range-value" for="${SLIDER_ID}"></output> ft`);
```
danielnaab: I don't think I've never seen the output tag used before! Why use `html()` rather than append a node?

assets/src/scripts/components/floodSlider.js
```diff
@@ -0,0 +1,60 @@
...
+    sliderContainer.append('input')
+        .attr('type', 'range')
+        .attr('class', 'wdfn-slider')
+        .attr('id', SLIDER_ID)
+        .on('change', dispatch(function() {
```
danielnaab: Would `oninput` work here, so it would update while dragging?

assets/src/scripts/components/floodSlider.js
```diff
@@ -0,0 +1,60 @@
...
+        const heightIndex = stages.indexOf(gageHeight);
+        node.select('input')
+            .attr('min', 0)
+            .attr('max', stages.length - 1)
+            .attr('step', 1)
+            .attr('value', heightIndex)
+            .attr('aria-valuemin', 0)
+            .attr('aria-valuemax', stages.length - 1)
+            .attr('aria-valuenow', heightIndex)
+            .attr('aria-valuetext', gageHeight);
```
danielnaab: I wonder if there are ways to provide useful information from a map to a screen reader? Would seem to be difficult..

assets/src/scripts/components/floodSlider.js
mbucknell: I hadn't either until I looked at how to make a slider accessible and there example used an output tag for the value. On the question, I just wanted the value in the output tag and wasn't sure how to get the text around it. Any suggestions?

assets/src/scripts/components/map.js
```diff
@@ -1,26 +1,122 @@
...
+            map.fitBounds(Util.extentToBounds(extent));
+        }
+    };
+
     // Add a gray basemap layer
     map.addLayer(new BasemapLayer('Gray'));
 
     // Add the ESRI World Hydro Reference Overlay
-    map.addLayer(new TiledMapLayer({url: HYDRO_URL}));
+    if (window.HYDRO_URL) {
```
danielnaab: Was this intended to be `window.HYDRO_ENDPOINT`? Is there a scenario we'd actually need this to be configurable? If not it seems unnecessary to make the Python-side aware of it.

Other thoughts... If we need to, we could consider using build-time variables for this sort of thing. `envify` is one way: https://github.com/hughsk/envify Perhaps if we need to share values between the two we could put them in JSON, and load it from config.py.

assets/src/scripts/components/floodSlider.js
mbucknell: I'll give it a try. I'm a little concerned that it would update the flood layers on the map too much.

assets/src/scripts/components/floodSlider.js
mbucknell: I have been thinking about this too but haven't done any research.

assets/src/scripts/components/map.js
mbucknell: It should be HYDRO_ENDPOINT. I did this for a couple of reasons. It would be good to put the configuration in one place and I wanted to be able to disable it for testing so that it doesn't create the layer (creating the layer would fetch the image then).

assets/src/scripts/components/floodSlider.js
danielnaab: Oh, now that you say that, I don't know... I'd guess `text()` but haven't tried.

assets/src/scripts/components/map.js
mbucknell: I'll look into envify, but I was following the pattern that we already have since many of our endpoints are only used by the javascript.

assets/src/scripts/components/floodSlider.js
mbucknell: Looks cool and there are a smallish number of stages typically so I'll make that change.

assets/src/scripts/store.js
```diff
@@ -103,13 +127,26 @@ export const Actions = {
...
 export const timeSeriesReducer = function (state={}, action) {
     let newState;
     switch (action.type) {
+        case 'SET_FLOOD_FEATURES':
+            return {
+                ...state,
+                floodStages: action.stages,
+                floodExtent: action.extent,
+                gageHeight: action.stages.length > 0 ? action.stages[0] : null
+            };
```
danielnaab: It would be nice to have a root reducer with separate reducers for each component - would that be hard to do as part of this story?

mbucknell: I think we should open a new ticket for that. I've spent a lot of time on this ticket already

assets/src/scripts/store.js
```diff
@@ -56,6 +58,28 @@ export const Actions = {
             );
         };
     },
+    retrieveFloodData(siteno) {
+        return function(displatch) {
```
danielnaab: typo, but displatch sounds like it should be a word :)

PR #124, opened 2018-03-12
Updated bumpversion to take a search strin for package.json and elimi…

mbucknell:
…nated package-lock.json from the file.

According to https://github.com/peritus/bumpversion/issues/149 , bumpversion does not handle multiline parsing when there are spaces as there would be in package-lock.json. We may need to consider other ways of keeping package-lock.json in sync.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -21,6 +21,6 @@ values =
 [bumpversion:file:code.json]
 
 [bumpversion:file:package.json]
-
-[bumpversion:file:package-lock.json]
+search = "version": "{current_version}"
+replace = "version": "{new_version}"
 


.bumpversion.cfg
```diff
@@ -21,6 +21,6 @@ values =
 [bumpversion:file:code.json]
 
 [bumpversion:file:package.json]
-
-[bumpversion:file:package-lock.json]
+search = "version": "{current_version}"
```
ayan-usgs: Wouldn't this still potentially be problematic after releases? For example, if a javascript package was also at version 0.4.0 or something similar.

I'm wondering if a multiline specification of the search string would give better specificity?

mbucknell: That line applies to package.json. I have removed package-lock.json from bumpversion. In package.json, only the package meta data has version: {version}

ayan-usgs: Ah. Yes. Should have looked more closely.

PR #126, opened 2018-03-12
WDFN-58 - Embed iframe snippet

danielnaab:


assets/src/scripts/components/embed.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+function attachToNode(store, node) {
+    // Select all text in input when it gets focus
+    const input = select(node).select('input');
+    input.on('focus', function () {
+        this.setSelectionRange(0, this.value.length);
+    });
+
+    // Prepend a font awesome icon to the beginning of the node
+    select(node).insert('a', 'fieldset');
```
mbucknell: Why not put this in the template?

assets/src/scripts/components/embed.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    });
+
+    afterEach(() => {
+        select('#component').remove();
+    });
+
+    it('should prepend sharing icon', () => {
+        expect(select('.fa-share').size()).toBe(1);
+    });
+});
```
mbucknell: We might want to consider testing of event handlers. I have artificially generated events in Jasmine tests: https://github.com/NWQMC/WQP_UI/blob/01863fd28fadbadf8de5fb094cc51754ea09429b/test/js/portal_ui/leafletControls/FeatureSourceSelectControlSpec.js#L73

waterdata/templates/monitoring_location.html
```diff
@@ -46,12 +48,7 @@ <h1>{{ stations[0].station_nm }}</h1>
                     <li><a href="{{ config.NWIS_ENDPOINTS.UV }}/?cb_00060=on&format=gif_mult_parms&site_no={{ stations[0].site_no }}&period=7" target="_blank">Interactive grapher</a></li>
                     <li><a href="{{ config.NWIS_ENDPOINTS.UV }}/?format=img_default&site_no={{ stations[0].site_no }}&period=7" target="_blank">Presentation Charts</a></li>
                 </ul>
-                <div class="wdfn-component" data-component="hydrograph" data-siteno="{{ stations[0].site_no }}">
-                    <div class="compare-container" data-siteno="{{ stations[0].site_no}}">
-                        <input ga-on="click" ga-event-category="TimeseriesGraph" ga-event-action="toggleCompare" id="show-last-year-input" class="hydrograph-last-year-input" type="checkbox" value="show-last_year" />
-                        <label for="show-last-year-input">Show last year</label>
-                    </div>
-                </div>
+                {{ components.TimeseriesComponent(stations[0].site_no) }}
```
mbucknell: I'd be okay with putting the floodSlider and map in a macro as you have done for the timeseries graph

assets/src/scripts/components/embed.js
danielnaab: I was shooting for progressive behavior, where the input is visible when javascript is disabled.

waterdata/templates/monitoring_location.html
danielnaab: OK

assets/src/scripts/components/embed.js
mbucknell: I must be missing something, but won't if the input is not visible then the icon would also be not visible?

waterdata/templates/partials/social_share.html
```diff
@@ -1,10 +1,17 @@
...
         <a href="https://www.facebook.com/sharer/sharer.php?u={{ share_encoded_url }}" target="_blank">
             <i class="fa fa-facebook-square" aria-label="Share on Facebook"></i>
         </a>
         <a href="https://twitter.com/home?status={{ share_encoded_url }}" target="_blank">
             <i class="fa fa-twitter-square" aria-label="Share on Twitter"></i>
         </a>
+        <span class="wdfn-component" data-component="embed">
+            <fieldset class="no-script">
+                <legend>Embed Snippet</legend>
+                <input type="text" id="embed-url" value='<iframe src="{{ url_for('time_series_component', _external=True, site_no=site_no) }}" width="600" height="900" frameborder="0" style="border:0" allowfullscreen></iframe>' aria-label="Embed this time-series chart on your own site" readonly>
```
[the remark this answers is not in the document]
danielnaab: @mbucknell aria-label is specified, which takes precedence over aria-labelledby. Is there an actual error here, or is this an accesslint issue?

assets/src/scripts/components/embed.js
danielnaab: I added the class `no-script` to the markup, which will set it to visible when Javascript is disabled. There's also a `has-script` class for the opposite condition.

waterdata/templates/partials/social_share.html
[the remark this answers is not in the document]
mbucknell: Sorry, I missed it because the line is so long... my bad.

assets/src/scripts/components/embed.js
mbucknell: Okay, I see what you are doing now.

PR #127, opened 2018-03-12
Wdfn 130 linear scale for everything but discharge

ayan-usgs:


assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -118,7 +134,16 @@ const xScaleSelector = memoize(tsKey => createSelector(
 const yScaleSelector = createSelector(
     layoutSelector,
     visiblePointsSelector,
-    (layout, pointArrays) => createYScale(pointArrays, layout.height - (MARGIN.top + MARGIN.bottom))
+    currentVariableSelector,
+    (layout, pointArrays, currentVar) => {
+        let currentVarParmCd;
+        try {
```
mbucknell: I prefer the tertiary operator here (at least I think this would work): ```currentVarParmCd = currentVarParmCd.variableCode ? currentVarParmCd.variableCode.value : null;```

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -43,6 +43,26 @@ function createXScale(values, xSize) {
...
 
+/**
+ * Create the scale based on the parameter code
+ *
+ * @param parmCd
+ * @param extent
+ * @param size
+ */
+function yScaleByParameter(parmCd, extent, size) {
+    const symlogParams = ['00060'];
```
mbucknell: Should this be a module level constant? I would prefer it use all caps at the very least.

ayan-usgs: Sure. It can be a module level constant.

assets/src/scripts/components/hydrograph/scales.js
ayan-usgs: I think that would work.

ayan-usgs: Turns out that doesn't actually work because the `currentVariableSelector` can return null, but this works: `currentVar && currentVar.variableCode ? currentVar.variableCode.value : null`.

PR #131, opened 2018-03-13
Wdfn 119 reformat tooltip text

ayan-usgs:


assets/src/scripts/components/hydrograph/tooltip.js
```diff
@@ -111,10 +111,17 @@ const updateTooltipText = function(text, {datum, qualifiers, unitCode}) {
             return;
         }
         let tzAbbrev = datum.dateTime.toString().match(/\(([^)]+)\)/)[1];
+        const maskKeys = new Set(Object.keys(MASK_DESC));
```
mbucknell: Does this need to go into a Set? By definition, the keys are all unique and you can use the function .includes on an array to test for membership.

assets/src/scripts/components/hydrograph/tooltip.js
```diff
@@ -111,10 +111,17 @@ const updateTooltipText = function(text, {datum, qualifiers, unitCode}) {
...
+        const qualiferKeysLower = new Set(datum.qualifiers.map(x => x.toLowerCase()));
```
mbucknell: Same question for keyInterset, does it need to be a Set?

assets/src/scripts/components/hydrograph/tooltip.js
ayan-usgs: Not strictly. I just wanted to a Set because I wanted to get the intersection between a data points qualifiers and masking qualifiers. There are other ways to do it, but it seemed logical in my mind.

PR #140, opened 2018-03-15
WDFN-111 Fix chart time range to IV request period

danielnaab:


assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -5,7 +5,7 @@ const { createSelector } = require('reselect');
 
 const { default: scaleSymlog } = require('../../lib/symlog');
 const { layoutSelector, MARGIN } = require('./layout');
-const { flatPointsSelector, timeSeriesSelector, variablesSelector, visiblePointsSelector, currentVariableSelector } = require('./timeseries');
+const { flatPointsSelector, timeSeriesSelector, variablesSelector, visiblePointsSelector, currentVariableSelector, requestTimeRangeSelector } = require('./timeseries');
```
mbucknell: flatPointsSelector is no longer used in scales.js

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -176,15 +175,16 @@ const parmCdTimeSeriesSelector = memoize(tsKey => createSelector(
  */
 const timeSeriesScalesByParmCdSelector = memoize(tsKey => memoize(dimensions => createSelector(
     parmCdTimeSeriesSelector(tsKey),
-    (timeSeriesByParmCd) => {
+    requestTimeRangeSelector,
+    (timeSeriesByParmCd, requestTimeRanges) => {
         return Object.keys(timeSeriesByParmCd).reduce((tsScales, parmCd) => {
             const seriesList = timeSeriesByParmCd[parmCd];
             const allPoints = seriesList.reduce((points, series) => {
```
mbucknell: Looks like allPoints could be removed.

assets/src/scripts/schema.js
```diff
@@ -72,7 +68,51 @@ const timeSeries = memoize(tsKey => new schema.Entity('timeSeries', {
...
+                notes[note.title] = note.value;
+                return notes;
+            }, {})
+        };
+        delete queryInfo.note;
+
+        // Extract values out of filter:timeRange
+        // If this is a "period" query (eg, P7D)
+        if (queryInfo.notes['filter:timeRange'].indexOf('PERIOD') > -1) {
+            const regEx = /\[mode=(.+), period=P(.+)D, modifiedSince=(.+)\]/;
```
mbucknell: It's too bad that this info is sent as a string containing an array rather than an array 😞

assets/src/scripts/components/hydrograph/scales.js
danielnaab: 👍 Thanks - my eslint plugin has stopped working.

PR #141, opened 2018-03-15
Wdfn 128 do not show table headers if no data

ayan-usgs:


assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -83,119 +83,121 @@ export const addSparkLine = function(svgSelection, {seriesLineSegments, scales})
  */
 export const plotSeriesSelectTable = function (elem, {availableTimeseries, lineSegmentsByParmCd, timeSeriesScalesByParmCd, layout}) {
     elem.select('#select-timeseries').remove();
-    const screenSizeCheck = layout.windowWidth <= SMALL_SCREEN_WIDTH;
```
danielnaab: Personally, to avoid extra indentation levels, I'd do this... but, tomato, tomahto. :)
```
if (!availableTimeseries.length) {
    return;
}
```

ayan-usgs: I like that.

PR #142, opened 2018-03-15
Wdfn 146 unavailable timeseries

ayan-usgs:


assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -24,7 +24,13 @@ export const availableTimeseriesSelector = createSelector(
 
         const codes = {};
         const seriesList = Object.values(timeSeries);
+        const timeSeriesVariables = seriesList.map(x => x.variable);
         for (const variableID of Object.keys(variables).sort()) {
+            // start the next iteration if a variable is not a
+            // series returned by the allTimeSeriesSelector
+            if (!timeSeriesVariables.includes(variableID)) {
```
danielnaab: I don't think we will currently polyfill the `includes` function, but we can deal with that once we get to old browser testing.

ayan-usgs: Roger that. There's also the option of falling back to the old-school `indexOf` method, but I'll leave it as is for now.

PR #143, opened 2018-03-15
WDFN-101 Slider

danielnaab:
This is a WIP and only includes a preliminary slider that updates the current tooltip date in the state.


assets/src/scripts/components/hydrograph/cursor.js
```diff
@@ -0,0 +1,43 @@
...
+                    // Use the integer time values of the x-axis range
+                    const domain = xScale.domain();
+                    const min = domain[0].getTime();
+                    const max = domain[1].getTime();
+                    input.attr('min', min)
+                        .attr('max', max)
+                        .attr('step', (max - min) / SLIDER_STEPS);
+                }, xScaleSelector('current')))
+                .call(link((input, cursorLocation) => {
+                    input.property('value', cursorLocation ? cursorLocation.getTime() : input.attr('max'))
```
mbucknell: Is 'value' an attribute?

assets/src/scripts/store.js
```diff
@@ -206,7 +212,14 @@ export const timeSeriesReducer = function (state={}, action) {
                     ...state.tooltipFocusTime,
```
mbucknell: Will tooltipFocusTime disappear eventually?

assets/src/scripts/components/hydrograph/cursor.js
danielnaab: Nope, it only worked as a property.

PR #147, opened 2018-03-16
Wdfn 118 disable last year checkbox

ayan-usgs:


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -269,8 +269,31 @@ const plotSROnlyTable = function (elem, {tsKey, variable, methods, visible, data
...
+            .dispatch('change'); // trigger a change event
+    } else {
+        checkbox.property('disabled', null);
+    }
+};
+
+
 const timeSeriesGraph = function (elem) {
-    elem.append('div')
+    elem.call(link(controlLastYearSelect, createStructuredSelector({
```
ayan-usgs: Opps.

ayan-usgs: Wait... I just had a brain-fart. I did want to do that.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -269,8 +269,31 @@ const plotSROnlyTable = function (elem, {tsKey, variable, methods, visible, data
...
+    }
+    const variableParmCd = variable.variableCode.value;
+    const availableVariableTimeseries = availableTimeseries.filter(x => x[0] === variableParmCd); // an array of arrays
+    const variableTimeseries = availableVariableTimeseries.map(x => x.slice(1));
+    const compareTimeseriesCountVals = variableTimeseries.map(x => x.map(y => y.compareTimeseriesCount));
+    if (compareTimeseriesCountVals.reduce((a, b) => Number(a) + Number(b), 0) === 0) {
+        checkbox
+            .property('disabled', true)
+            .property('checked', false)
+            .dispatch('change'); // trigger a change event
```
ayan-usgs: I'm on the fence about whether or not this is confusing behavior.... :thinking:

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -269,8 +269,31 @@ const plotSROnlyTable = function (elem, {tsKey, variable, methods, visible, data
     }
 };
 
+const controlLastYearSelect = function(elem, {variable, availableTimeseries}) {
+    let checkbox = elem.select('.hydrograph-last-year-input');
+    if (variable === null || !availableTimeseries.length) {
```
mbucknell: Wouldn't the control be disabled if the time series length was zero?

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -269,8 +269,31 @@ const plotSROnlyTable = function (elem, {tsKey, variable, methods, visible, data
...
+    } else {
+        checkbox.property('disabled', null);
+    }
+};
+
+
 const timeSeriesGraph = function (elem) {
-    elem.append('div')
+    elem.call(link(controlLastYearSelect, createStructuredSelector({
+        variable: currentVariableSelector,
```
mbucknell: So this is going to conflict with the work I'm doing now. We will no longer be showing the availablility of the time series in the parameters list, so I had changed the selector to no longer gather the counts.. I wonder if we could just make this simpler, by using the currentVariableTimeSeriesSelector('compare') (in the timeseries module) and seeing if the length of the timeseries is zero?

ayan-usgs: That would work. I'll make that change.

PR #148, opened 2018-03-19
Wdfn 122 - Refine current conditions list

mbucknell:


assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -4,7 +4,7 @@ const { select } = require('d3-selection');
 
 const { allTimeSeriesSelector } = require('./timeseries');
 const { Actions } = require('../../store');
-const { SPARK_LINE_DIM, SMALL_SCREEN_WIDTH } = require('./layout');
+const { SPARK_LINE_DIM } = require('./layout');
```
ayan-usgs: I think this was the only place where `SMALL_SCREEN_WIDTH` is being used. Perhaps we can remove it?

PR #149, opened 2018-03-19
Wdfn 123 sort iv data in relevant way

ayan-usgs:


assets/src/scripts/models.js
```diff
@@ -10,6 +10,13 @@ const { deltaDays } = require('./utils');
 const SERVICE_ROOT = window.SERVICE_ROOT || 'https://waterservices.usgs.gov/nwis';
 const PAST_SERVICE_ROOT = window.PAST_SERVICE_ROOT  || 'https://nwis.waterservices.usgs.gov/nwis';
 
+export const PARAM_PERTINENCE = {
+    '00060': {'rank': 0},
+    '00065': {'rank': 1},
+    '72019': {'rank': 2}
+};
```
danielnaab: Could you remove the object containing rank here, and just store the integer?

ayan-usgs: Yep.

PR #153, opened 2018-03-20
Wdfn 144 - Add a title to the graph

mbucknell:
Ended up adding a div above the svg for the title. This way it doesn't interfer with the tooltip text.

### DIFF
--- a/assets/src/scripts/components/hydrograph/index.js
+++ b/assets/src/scripts/components/hydrograph/index.js
@@ -289,6 +289,22 @@ const controlLastYearSelect = function(elem, {compareTimeseries}) {
     }
 };
 
+const createTitle = function(elem) {
+    elem.append('text')
+        .classed('timeseries-graph-title', true)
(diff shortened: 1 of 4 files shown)


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -289,12 +289,20 @@ const controlLastYearSelect = function(elem, {compareTimeseries}) {
     }
 };
 
+const createTitle = function(elem) {
```
ayan-usgs: I'm wondering if it would be more consistent with the rest of the module if this function took the `titleSelector` as an argument instead. But whatever, it's six of one and half dozen of the other.

mbucknell: Several use this same pattern (e.g., createTooltipText). I want the location of this div to only be rendered once, so there are no surprises as to where it is in the markup. Doing this way means only the contents of the div change.

PR #154, opened 2018-03-20
Wdfn 150 do not show graph for no data sites

ayan-usgs:
I also tried dealing with not showing the graph or any of its peripheral controls by not appending any thing to div if there aren't timeseries, but that got real messy. I opted to hide the plot by adjusting the styling.

### DIFF
--- a/assets/src/scripts/components/hydrograph/index.js
+++ b/assets/src/scripts/components/hydrograph/index.js
@@ -20,7 +20,7 @@ const { Actions } = require('../../store');
 const { pointsSelector, pointsTableDataSelector, lineSegmentsByParmCdSelector, currentVariableLineSegmentsSelector,
     MASK_DESC, HASH_ID } = require('./drawingData');
 const { currentVariableSelector, methodsSelector, isVisibleSelector, titleSelector,
-    descriptionSelector,  currentVariableTimeSeriesSelector, allTimeSeriesSelector } = require('./timeseries');
+    descriptionSelector,  currentVariableTimeSeriesSelector, timeSeriesSelector } = require('./timeseries');
 const { createTooltipFocus, createTooltipText } = require('./tooltip');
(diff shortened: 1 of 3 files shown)


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -296,9 +296,31 @@ const createTitle = function(elem) {
...
         }, titleSelector));
 };
+/**
+ * Modify styling to hide or display the plot area.
+ *
+ * @param elem
+ * @param allTimeseries
+ */
+const controlGraphDisplay = function (elem, allTimeseries) {
+    const seriesWithPoints = Object.values(allTimeseries).filter(x => x.points.length > 0 && x.tsKey === 'current');
```
mbucknell: You can at least partially filter the time series with the timeSeriesSelector('current') (in the timeseries module).

waterdata/templates/macros/components.html
```diff
@@ -1,6 +1,6 @@
 {% macro TimeseriesComponent(site_no) -%}
     <div class="wdfn-component" data-component="hydrograph" data-siteno="{{ site_no }}">
-        <div class="compare-container" data-siteno="{{ site_no }}">
+        <div class="compare-container" style="display: none;" data-siteno="{{ site_no }}">
```
mbucknell: I wonder if using the "hidden" attribute would work? That's the recommendation from the designsystem.digital.gov website.

ayan-usgs:
> check to see if the hidden attribute would work

`hidden` works pretty well.

assets/src/scripts/components/hydrograph/index.js
ayan-usgs: Good idea! There are so many different selectors, I sometimes lose track of them. :flushed:

PR #159, opened 2018-03-22
provide error handling for bbox failures when the div isn't being shown

ayan-usgs:


ayan-usgs:
> I wonder if we should reconsider not drawing the svg at all if there is no data. I know that there are implications to that and it would actually be a bigger change. Just a thought.

I think that's worth attempting again. It was pretty difficult to get the SVGs and it's associated controls to render correctly when the data becomes available because the redux state doesn't have data initially and then it would about 100 ms later. I up for poking at that some, though I might merge this in before attempting it.

PR #161, opened 2018-03-22
Wdfn 143 hide troubleshooting links

ayan-usgs:


config.py
```diff
@@ -9,6 +9,10 @@
 
 from waterdata import __version__ as APP_VERSION
 
+# controls environment specific behavior
+# acceptable values: 'local', 'development', 'staging', 'production'
+DEPLOYMENT_ENVIRONMENT = 'development'
```
mbucknell: I would suggest we set this to "prod". It's the safer thing to do if for some reason, the config is not updated correctly. Also the test in monitoring_location is for the string "prod", so the comment needs to be updated.

ayan-usgs: :man_facepalming:

PR #164, opened 2018-03-23
Updated change log.

mbucknell:


mbucknell:
Please let me know if I'm missing something from the Changelog.

ayan-usgs:
Hmm... not sure if these two items count as changes or additions (or if they should be included at...):

- Graphs are not shown for sites with not current times
- Parameters with no data points in current or last year data are omitted from the select table

ayan-usgs:
Line 21 appears to be missing a bullet.

ayan-usgs:
Also, graphs are not shown for sites with no current data.

mbucknell:
Regarding Line 21. That's correct. Two things were done to the select a timeseries list visually.

PR #165, opened 2018-03-23
Tooltips for all time series - WDFN-154

danielnaab:
This displays the tooltip for all visible time series, and also fixes an issue introduced with the slider ticket where the tooltip mouse events were being intercepted by elements above them in the DOM.

### DIFF
--- a/assets/src/scripts/components/hydrograph/tooltip.js
+++ b/assets/src/scripts/components/hydrograph/tooltip.js
@@ -96,7 +96,9 @@ const createTooltipTextGroup = function (elem, {currentPoints, comparePoints, qu
     // Put the circles in a container so we can keep the their position in the
     // DOM before rect.overlay, to prevent the circles from receiving mouse
     // events.
+    let resizeBackground = true;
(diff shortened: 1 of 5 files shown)


assets/src/scripts/components/hydrograph/tooltip.js
```diff
@@ -64,13 +64,9 @@ const tooltipPointsSelector = memoize(tsKey => createSelector(
     }
 ));
 
-const updateTooltipText = function(text, {cursorPoints, qualifiers, unitCode}) {
+const getTooltipText = function(datum, qualifiers, unitCode) {
```
ayan-usgs: Would it be worthwhile to add a test for this function? It's previous variant didn't have a test, so I think it's a show stopper, but I think it would be good to add tests for functions if we're touching them and they don't already have them.

danielnaab: I'm inclined to keep it as-is since the code is being tested higher up the chain.

PR #168, opened 2018-03-26
Wdfn 140 - Move graph controls to legend area

mbucknell:


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -162,7 +163,8 @@ const plotSvgDefs = function(elem) {
 
 const timeSeriesLegend = function(elem) {
     elem.append('div')
-        .attr('class', 'hydrograph-container')
+        .classed('timeseries-legend-container', true)
```
danielnaab: I don't this class is used anywhere

mbucknell: Yeah, I was using it in the css but I am not anymore.

danielnaab:
Feels much nicer with the buttons on the bottom

PR #169, opened 2018-03-26
Visual Identity WDFN-139

danielnaab:
Logo sizes taken from usgs.gov, but spread over the four UWSDS breakpoints.


waterdata/templates/base_plain.html
```diff
@@ -21,5 +22,36 @@
         <script>document.body.className += ' js';</script>
         {% block body %}{% endblock %}
     </body>
+    {% if config.LIVE_RELOAD_PATH %}
```
mbucknell: 👍

PR #176, opened 2018-03-28
update change log with 0.4.0's release date and start an unreleased s…

ayan-usgs:
…ection


ayan-usgs:
Not sure if this the best moment to update the change log, but it looked like version 0.4.0 wasn't identified in the log.

mbucknell:
My bad. I did an incomplete update before making the release. I think its fine to merge.

PR #177, opened 2018-03-28
Wdfn 183 - Separate feature toggles for the Audio and Hydrologic

mbucknell:
This also includes code to not bring in the AudioContext if not enabled in the audible module. Jim tested this and it seems to eliminate the bug he was seeing.

### DIFF
--- a/assets/src/scripts/components/hydrograph/audible.js
+++ b/assets/src/scripts/components/hydrograph/audible.js
@@ -46,28 +46,30 @@ export const createSound = memoize(/* eslint-disable no-unused-vars */ tsKey => 
 });
 
 export const updateSound = function ({enabled, points}) {
-    const audioCtx = getAudioContext();
-    for (const tsKey of Object.keys(points)) {
-        const point = points[tsKey];
-        const {compressor, oscillator, gainNode} = createSound(tsKey);
+    if (enabled) {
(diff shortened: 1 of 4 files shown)


assets/src/scripts/components/hydrograph/audible.js
```diff
@@ -46,28 +46,30 @@ export const createSound = memoize(/* eslint-disable no-unused-vars */ tsKey =>
...
-            audioCtx.currentTime,
-            .2
-        );
-
-        gainNode.gain.setTargetAtTime(
-            enabled && point ? volumeScale(point) : null,
-            audioCtx.currentTime,
-            .2
-        );
+    if (enabled) {
```
danielnaab: I think this might prevent the sound from being turned off if the audible checkbox is clicked while the sound is playing. Could you check if AudioContext == null instead?

mbucknell: I put this in to prevent the sound from playing on Firefox and Safari when the page was loaded. When clicking the toggle, the sound does stop.

danielnaab: Huh - I'm surprised, but sounds good.

danielnaab:
👍

PR #179, opened 2018-03-28
Hash assets WDFN-152

danielnaab:


assets/gulpfile.js
```diff
@@ -0,0 +1,17 @@
...
+
+
+gulp.task('revall', function () {
+    return gulp.src('dist/**')
+        .pipe(RevAll.revision({
+            dontGlobal: [
+                /rev-manifest.json$/,
+                /.+\.[a-f0-9]{8}\..+$/  // skip already hashed files
+            ],
+            includeFilesInManifest: ['.css', '.js', '.ico', '.svg', '.png']
```
danielnaab: This is the one annoyance with gulp-rev-all... we can't tell it to include all files in the manifest. So if we need other extensions on the Python-side, we'll need to add them here.

PR #182, opened 2018-03-28
Wdfn 142 character description of a site

ayan-usgs:


waterdata/location_utils.py
```diff
@@ -312,3 +313,74 @@ def parm_grp_sort(x):
...
+    all_parms = list(itertools.chain.from_iterable([x['parameters'] for x in rolled_up_dataseries.values()]))
+    all_start_dates = [parm['start_date'] for parm in all_parms]
+    try:
+        start_year = min(all_start_dates).year
+    except ValueError:
+        start_year = None
+    # need to use a set -- there might be multiple variants for a measurements
+    # (e.g. "nitrite, water as N" vs "nitrite, water as NO3-")
+    short_parm_names = set([parm['parameter_name'].split(',')[0].upper() for parm in all_parms])
+    sort_top = ('DISCHARGE', 'GAGE HEIGHT', 'TEMPERATURE')
```
mbucknell: Consider using all caps for the variable name

waterdata/location_utils.py
```diff
@@ -312,3 +313,74 @@ def parm_grp_sort(x):
...
+    sorted_parm_names = sorted(short_parm_names, key=lambda x: (x not in sort_top, x))
```
mbucknell: I see why this works. It will not work if we want TEMPERATURE before GAGE_HEIGHT for instance, but as long as it doesn't matter this works.

waterdata/location_utils.py
```diff
@@ -312,3 +313,74 @@ def parm_grp_sort(x):
...
+    :param str county: country where the monitoring location lives
+    :param str state: state where the monitoring location lives
+    :param dict rolled_up_dataseries: data series rolled-up by parameter group with certain groups we don't care about
+        excluded
+    :return: description text for a site
+    :rtype: str
+
+    """
+    site_type_article = use_correct_indefinite_article(location_type)
+    desc = 'Monitoring location {loc_id} is associated with {article} {loc_type} in {county}, {state}.'.format(
```
mbucknell: I wonder if this can be placed in the template. The use_correct_indefinite_article could be a filter. You could pass the parameters and start_date into the template. We might be able to use a macro so that it isn't duplicated for both the header tag and the page text.

waterdata/templates/base_plain.html
```diff
@@ -5,6 +5,7 @@
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta name="format-detection" content="telephone=no">
+        {% block description %}{% endblock %}
```
mbucknell: Consider using a more general term. This block could be used to any number of meta tags.

waterdata/location_utils.py
ayan-usgs: I of thinking that I don't care about the order of those three items as long as they take priority over things like "Agency analyzing sample".

ayan-usgs:
> My biggest concern is that we are doing two much formatting in python rather than the Jinja templates

Yeah... I was thinking about that, but I felt like putting it in the python was more testable, particularly for grammatical nuances that were happening based on list size. Regardless, I'm up for fiddling with shoving more stuff in Jinja.

mbucknell:
Another Jinja filter for pluralization?!

danielnaab:
Django has one that might be a starting point: https://github.com/django/django/blob/5033999153f7303a18d94ea2f08ed78545c0f3df/django/template/defaultfilters.py#L852

ayan-usgs:
> Another Jinja filter for pluralization?!

Maybe. Though I was actually more thinking long the lines of stuff like: `parm1` vs `parm1 and parm2` vs `parm1, parm2, and parm 3` vs `parm1, parm2, parm3, and more`. But that could probably be a filter too.

ayan-usgs:
Moved most of the description code into the templates and filters.

PR #191, opened 2018-03-30
Wdfn159 Crawlable County and State Pages

abriggs-usgs:


waterdata/tests/test_views.py
```diff
@@ -149,6 +149,39 @@ def test_locations_list(self, client):
         assert text.count('01630500') == 16, 'Expected site 01630500 in output'
 
 
+class TestCountryStateCountyView:
+    @pytest.fixture(autouse=True)
```
mbucknell: If you add ```# pylint: disable=R0201``` above this line you will disable some pylint errors.

waterdata/templates/states_counties.html
```diff
@@ -0,0 +1,89 @@
...
+            {% if show_locations_link %}
+                 <a href="{{ url_for('county_station_locations', state_cd=state_cd, county_cd=county_cd) }}">Monitoring Locations</a>
+            {% endif %}
+
+
+        {% else %}
+            <h1>Error: HTTP {{ http_code }} -- State or county not found.</h1>
+        {% endif %}
+
+        {% if monitoring_locations %}
```
mbucknell: I would put this before the else if line 63 to make clear that it's only shown if it's not a 404.

waterdata/views.py
```diff
@@ -206,6 +208,67 @@ def hydrological_unit_locations(huc_cd):
...
+        unit_data = app.config['COUNTRY_STATE_COUNTY_LOOKUP']['US']['state_cd'].get(state_cd, None)['county_cd'].get(county_cd, None)
+
+    # Get the data corresponding to this state
+    if state_cd and not county_cd:
+        unit_data = app.config['COUNTRY_STATE_COUNTY_LOOKUP']['US']['state_cd'].get(state_cd, None)
+
+    # If no state and or state and county code is available, display list of states.
+    if not state_cd and not county_cd:
+        unit_data = {
+            'name': 'US',
```
mbucknell: Consider assigning the name to be 'United States' rather than having an if statement in the template.

PR #196, opened 2018-04-02
Wdfn 161 - Connect current gage height to FIM slider

mbucknell:


assets/src/scripts/store.js
```diff
@@ -7,11 +9,23 @@ const { getMedianStatistics, getPreviousYearTimeseries, getTimeseries,
...
+        return ts.variable === parmVar;
+    });
+    let points = parmTimeSeries ? collection.timeSeries[parmTimeSeries].points : [];
+    return points.length ? last(points).value : null;
+};
+
+const GAGE_HEIGHT_CD = '00065';
 
 export const Actions = {
-    retrieveTimeseries(siteno, params=null, startDate=null, endDate=null) {
```
mbucknell: I removed these as parameters. The code at  line 35 and 36 which set the endTime and startTime assume that we have requested a period of time. This makes it explicit that this action gets the most recent data.

assets/src/scripts/store.js
```diff
@@ -158,14 +178,37 @@ export const Actions = {
...
+    },
+    setGageHeight(gageHeight) {
+        return {
+            type: 'SET_GAGE_HEIGHT',
+            gageHeight
+        };
     }
 };
 
+const setGageHeight = function(gageHeight, floodStages) {
```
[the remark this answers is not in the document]
mbucknell: Yes ... too many setGageHeights!

assets/src/scripts/store.js
```diff
@@ -25,6 +39,12 @@ export const Actions = {
                     // Trigger a call to get last year's data
                     dispatch(Actions.retrieveCompareTimeseries(siteno, startTime, endTime));
 
+                    // Update the gage height if gage height is in the data.
+                    const currentGageHeight = getLatestValue(collection, GAGE_HEIGHT_CD);
+                    if (currentGageHeight) {
+                        dispatch(Actions.setGageHeight(currentGageHeight));
+                    }
```
danielnaab: I'm wondering if this logic could be done in the flood slider component if `gageHeight === null` rather than here, to keep timeSeries handling separate from the gage height.

mbucknell: The problem as I see it is we have two ways of setting the gageHeight. One is retrieving the current time series and the other is with the slider. I'm concerned that if we don't do it as part of the time series retrieval, there will be some nasty conditions where the flood slider has been rendered and used before the timeseries data has finished loading.

PR #197, opened 2018-04-02
Wdfn 171 data mask labels for sparkline svg

ayan-usgs:


assets/src/scripts/components/hydrograph/parameters.js
```diff
@@ -64,12 +65,44 @@ export const addSparkLine = function(svgSelection, {seriesLineSegments, scales})
...
+                    .attr('x', 0)
+                    .attr('y', yPosition)
+                    .text(maskText);
+                centerElement(svgText);
+                centerElement(tspan);
+            });
+        } else {
+            svgText.text(maskDesc)
+                .attr('y', '20');
+            centerElement(svgText);
```
danielnaab: This looks good, so not suggesting we change it.. but just going through this with the tooltips, I think putting the text straight into the table as a div (or whatever) rather than drawing SVG could be simpler.

ayan-usgs: I think I might keep it as is for now. I kind of like how the text is currently confined to the same space as the other sparklines... seems visually consistent.

PR #198, opened 2018-04-03
Wdfn 198 uv time params in desc

ayan-usgs:


waterdata/filters.py
```diff
@@ -40,36 +42,45 @@ def use_correct_indefinite_article(some_noun):
...
     # need to use a set -- there might be multiple variants for a measurements
     # (e.g. "nitrite, water as N" vs "nitrite, water as NO3-")
+    series = chain.from_iterable(
+        [
+            x['parameters'] for x in parameter_group_series.values() if
+            'unit values' in x['data_types'].lower() and
+            (datetime.datetime.now().date() - datetime.timedelta(days=8) <=
+             x['end_date'].date() <= datetime.datetime.now().date() + datetime.timedelta(days=1))
+        ]
+    )
```
ayan-usgs: Mixed feelings about the date filtering here. There are `timedelta`s to handle instances if code is being executed around the site is in a different timezone.

In an ideal world, the comparison here would be done in UTC, but I couldn't think of a way to get a site's UTC offset without adding at least another service call. I wagered it wasn't worth it, since this will probably be okay most of the time.

waterdata/filters.py
```diff
@@ -40,37 +42,45 @@ def use_correct_indefinite_article(some_noun):
...
     """
     # need to use a set -- there might be multiple variants for a measurements
     # (e.g. "nitrite, water as N" vs "nitrite, water as NO3-")
-    short_names = set([s['parameter_name'].split(',')[0].upper() for s in series])
+    series = chain.from_iterable([x['parameters'] for x in parameter_group_series.values()])
+    short_names = set(
+        [
+            s['parameter_name'].split(',')[0].upper() for s in series if 'Unit Values' in s['data_types'] and
+            (datetime.datetime.now().date() - datetime.timedelta(days=8) <=
+             s['end_date'].date() <= datetime.datetime.now().date() + datetime.timedelta(days=1))
```
ayan-usgs: Mixed feelings about the date filtering here. There are timedeltas to handle instances if code is being executed around the site is in a different timezone.

Ideally, the comparison here would be done in UTC, but I couldn't think of a way to get a site's UTC offset without adding at least another service call. I wagered it wasn't worth it, since this will probably be okay most of the time.

mbucknell: I think this is fine for purposes of this ticket.

PR #199, opened 2018-04-03
WDFN-185 - Fix left margin sizing issues

danielnaab:
Move domain code out of scales.js into domain.js, so we can import it in layout.js without circular dependencies. Use the tick labels to calculate the necessary left margin size.

I'm not 100% happy with this, but it does fix the margin issue...

I had also looked into creating a margin.js module separate from layout.js, but that produces circular dependencies on axes.js.

If we moved the rendering code out of axes.js into index.js, that might move things a bit closer. However, we currently need both margin and layout info to create the scales, which would in turn be needed to create the margins.

Perhaps a proper solution includes making the scale ranges solely a function of the browser width (and not the margins). But that might have larger implications that I don't want to deal with as a part of this.

### DIFF
--- a/assets/src/scripts/components/hydrograph/domain.spec.js
+++ b/assets/src/scripts/components/hydrograph/domain.spec.js
@@ -1,4 +1,4 @@
-const { extendDomain } = require('./domain');
+const { extendDomain, getYDomain, getYTickDetails } = require('./domain');
 
 
 describe('domain module', () => {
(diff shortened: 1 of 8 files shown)


assets/src/scripts/components/hydrograph/layout.js
```diff
@@ -38,12 +41,19 @@ const SPARK_LINE_DIM = {
 const layoutSelector = createSelector(
     (state) => state.width,
     (state) => state.windowWidth,
-    (width, windowWidth) => {
+    tickSelector,
+    (width, windowWidth, tickDetails) => {
+        const margin = mediaQuery(USWDS_SITE_MAX_WIDTH) ? MARGIN : MARGIN_SMALL_DEVICE;
+        const tickLengths = tickDetails.tickValues.map(v => tickDetails.tickFormat(v).length);
+        const approxLabelLength = Math.max.apply(null, tickLengths) * 10;
```
ayan-usgs: I would have opted for `Math.max(...arr)` here, but tomayto, tomahto. :man_shrugging:

assets/src/scripts/components/hydrograph/domain.spec.js
```diff
@@ -0,0 +1,40 @@
+const { extendDomain } = require('./domain');
+
+
+describe('domain module', () => {
```
mbucknell: It would be great to add tests for getYDomain and getYTickDetails.

assets/src/scripts/components/hydrograph/layout.js
danielnaab: I had been erroring on the side of "performance", because I had read somewhere that the spread operator is expensive... but apparently that's only with Traceur, not Babel: http://incaseofstairs.com/2015/06/es6-feature-performance/ Might as well use the spread operator everywhere.

PR #205, opened 2018-04-04
Wdfn 199 single points among masks

ayan-usgs:


assets/src/scripts/components/hydrograph/timeseries.js
```diff
@@ -3,6 +3,13 @@ const memoize = require('fast-memoize');
 const { createSelector } = require('reselect');
 
 
+// array of parameters that should use
+// a symlog scale instead of a linear scale
+export const SYMLOG_PARMS = [
```
ayan-usgs: Noticed that this was duplicated in both scales and domain.js. I moved it to timeseries.js for lack of idea of where to put it.

mbucknell: It could be exported from domain as scales depends on domain. Just a thought

ayan-usgs: That's a good idea.

PR #209, opened 2018-04-05
WDFN-206 - Try to show fewer ticks on log scales with small screen widths

danielnaab:


assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -101,8 +105,23 @@ export const yDomainSelector = createSelector(
...
+        }
+        if (mediaQuery(config.USWDS_MEDIUM_SCREEN)) {
+            tickCount++;
+        }
+        if (mediaQuery(config.USWDS_LARGE_SCREEN)) {
+            tickCount++;
+        }
+    }
+
+    const tickValues = ticks(yDomain[0], yDomain[1], tickCount);
```
danielnaab: This ticks function treats the tickCount as a "suggestion", so it's sometimes unpredictable. This method of choosing the count is working for the test sites I've encountered, but does often lead towards only a top and bottom tick on small widths.

PR #210, opened 2018-04-05
Wdfn 212 - Precip is always zero

mbucknell:


assets/src/scripts/components/hydrograph/drawingData.js
```diff
@@ -96,6 +96,27 @@ export const pointsByTsKeySelector = memoize(tsKey => createSelector(
         return result;
     }));
 
+/* Returns a select that returns all time series point for the ccurrent variable and in the select series, tsKey
+ * by tsId.
+ * @param {Object} state
+ * @param {String} tsKey
+ * @return Object
+ */
+export const currentVariablePointsByTsIdSelector = memoize(tsKey => createSelector(
```
mbucknell: I added this because the tsPointsCursorSelector expects the data to be returned by tsId. I'm not sure this is necessary and will continue to look into whether it is, but wanted to get this out since it does fix the problem

mbucknell: Pretty sure it is necessary.

PR #211, opened 2018-04-05
Wdfn 178 - Accommodate showing a site that is actually outside of the FIM map

mbucknell:


mbucknell:
This included changes for another ticket so closing

PR #215, opened 2018-04-06
WDFN-206 - Take every other log scale tick

danielnaab:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-206
-----------
https://internal.cida.usgs.gov/jira/browse/WDFN-206

Description
-----------
Only use every other tick for log scales on small device widths when the tick count is 4 or more.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/domain.js
+++ b/assets/src/scripts/components/hydrograph/domain.js
@@ -11,7 +11,6 @@ const { visiblePointsSelector } = require('./drawingData');
 
 const PADDING_RATIO = 0.2;
 const Y_TICK_COUNT = 5;
-const Y_TICK_COUNT_SMALL_LOG = 2;
 // array of parameters that should use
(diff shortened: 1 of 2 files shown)


assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -107,21 +106,14 @@ export const yDomainSelector = createSelector(
...
-        if (mediaQuery(config.USWDS_MEDIUM_SCREEN)) {
-            tickCount++;
-        }
-        if (mediaQuery(config.USWDS_LARGE_SCREEN)) {
-            tickCount++;
-        }
+
+    let tickValues = ticks(yDomain[0], yDomain[1], Y_TICK_COUNT);
+
+    // On small screens, log scale ticks are too close together, so only use every other one.
```
mbucknell: This is fine, but I wonder if it could be done in the css?

PR #227, opened 2018-04-16
Wdfn85

abriggs-usgs:
Before making a pull request
----------------------------

- [X] Make sure all tests run

WDFN 85
-----------
Added the main meta tags required by Twitter Cards and Facebook Open Graph. The image meta tags requested in the ticket have not been added because SVG elements, like the graph, are not one of the formats currently accepted by Twitter Cards or Open Graph. 

After making a pull request
---------------------------

### DIFF
--- a/waterdata/templates/monitoring_location.html
+++ b/waterdata/templates/monitoring_location.html
@@ -6,6 +6,18 @@
 {% block extra_head_tags %}
     {% if status_code == 200 %}
         <meta name="description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}">
+                <!-- tags for Facebook Open Graph -->
+        <meta property="og:url" content="{{ url_for('monitoring_location', site_no=site_no, _external=True) }}" />
+        <meta property="og:type" content="website" />
+        <meta property="og:title" content="{{ page_title }}" />
+        <meta property="og:description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}" />
+        <meta property="og:image" content="<iframe src='{{ url_for('time_series_component', site_no=site_no, _external=True) }}' width='600' height='900' frameborder='0' style='border:0' allowfullscreen></iframe>" />
+        <!-- tags for Twitter Cards -->
+        <meta name="twitter:card" content="summary_large_image">
+        <meta name="twitter:site" content="@USGS">
+        <meta name="twitter:title" content="{{ page_title }}">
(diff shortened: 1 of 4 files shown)


waterdata/templates/monitoring_location.html
```diff
@@ -6,6 +6,16 @@
 {% block extra_head_tags %}
     {% if status_code == 200 %}
         <meta name="description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}">
+        <!-- tags for Facebook Open Graph -->
```
mbucknell: I think we shouldn't be adding any of these meta tags if more than one site has been passed in. So this will need an if clause similar to the one on line 36

waterdata/templates/monitoring_location.html
```diff
@@ -6,6 +6,16 @@
...
+        <meta property="og:url" content="{{ url_for('monitoring_location', site_no=site_no, _external=True) }}" />
+        <meta property="og:type" content="website" />
+        <meta property="og:title" content="{{ page_title }}" />
```
mbucknell: Does this work? I'm a little surprised because it's not set until line 31. If it is working, maybe move line 31 before this part of the code.

waterdata/templates/monitoring_location.html
```diff
@@ -6,6 +6,16 @@
 {% block extra_head_tags %}
     {% if status_code == 200 %}
         <meta name="description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}">
+        <!-- tags for Facebook Open Graph -->
+        <meta property="og:url" content="{{ url_for('monitoring_location', site_no=site_no, _external=True) }}" />
```
mbucknell: Can you use  stations[0].siteno here rather than explicitly passing in the site_no in the context?

waterdata/templates/monitoring_location.html
abriggs-usgs: Good point. I didn't think of that.

waterdata/templates/monitoring_location.html
```diff
@@ -6,6 +6,19 @@
 {% block extra_head_tags %}
     {% if status_code == 200 %}
         <meta name="description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}">
+        {% if stations|length == 1 %}
```
mbucknell: Line 8 should be within the if clause too.

PR #229, opened 2018-04-18
Wdfn 190 - Refactor redux reducers

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
WDFN 190 - Refactor redux reducers

Description
-----------
Changed the shape of the state into domain data (series and floodData), application state (timeseriesState and floodState) and ui state (ui). Each has it's own slice reducer and within the slice reducer, case reducer functions are used to implement each case.

After making a pull request
---------------------------


assets/src/scripts/store/seriesReducer.spec.js
```diff
@@ -0,0 +1,195 @@
...
+                    },
+                    compare: {
+                        queryInfo: 'compare'
+                    }
+                }
+            }, {
+                type: 'RESET_TIMESERIES',
+                key: 'compare'
+            });
+            console.log(result);
```
ayan-usgs: Should this console.log be here?

assets/src/scripts/store/index.js
```diff
@@ -192,166 +222,43 @@ export const Actions = {
...
-        statisticalMetaData: {
-            beginYear: '',
-            endYear: ''
+        floodData: {
+            stages: [],
+            extent: {}
+        },
+
+        timeseriesState: {
+            showSeries: {
```
ayan-usgs: I think all of these changes are logical and make sense in my head. Though, I seem to recall us making some effort to normalize the state and reduce nesting. That said, I can't think of a more logical to handle the various timeseries options, so this can probably be left as it is.

mbucknell: We did flatten 'series' when Dan brought in Normalizr. I saw this effort as more trying to organize  and shape the store to match the components on the page. I think we still need to work on perhaps refactoring the actions and then for sure when need to work on the selectors for the hydrograph. I'm actually right now reworking the flood data and map component but that is way easier than timeseries/hydrograph. I will use a separate PR for that.

assets/src/scripts/store/seriesReducer.spec.js
mbucknell: Nope. I will remove it in the next PR. I'm going to go ahead and merge this one.

mbucknell:
I will remove the console.log with my next PR for this ticket.

PR #230, opened 2018-04-18
Wdfn 190 part2

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run
- [X] Update the changelog appropriately - no changes needed

Title
-----------
Refactored the map component and incorporated the flood slider in the map component since it makes no sense to have these separate.

Description
-----------
I have been reading more about how to structure Redux applications. Currently, I don't like blending our selectors with our view code. The refactor of the map component shows a new module floodDataSelector. This contains selectors for each property of flood data (so stages, extent, and gageHeight) along with selectors that derive information. The map/index.js and floodslider.js only use selectors from this module, therefore they don't need to be aware of the shape of the state. I think we should consider a similar pattern for the hydrograph module. Obviously it would probably not be a single selector module but I could foresee potentially something like scaleSelector, timeseriesSelector, etc.

After making a pull request
---------------------------


assets/src/scripts/components/map/floodDataSelector.js
```diff
@@ -0,0 +1,55 @@
...
+            result = stages[0];
+            let diff = Math.abs(gageHeight - result);
+            stages.forEach((stage) => {
+                let newDiff = Math.abs(gageHeight - stage);
+                if (newDiff <= diff) {
+                    diff = newDiff;
+                    result = stage;
+                }
+            });
+        }
```
ayan-usgs: I've been pondering this function... do you think it'd be more efficient to do this to be the min?
```
result = stages.filter(i => Math.abs(gageHeight-i) == Math.min(...stages.map(i => Math.abs(gageHeight-i))))[0]
```

mbucknell: That's a great idea! Thanks

mbucknell: On second glance, I think I'll leave as is. I think that function is no more efficient although it is more concise. I found it a little hard to reason about.

PR #231, opened 2018-04-18
Making changes to fix bug that prevented name of monitoring site from…

abriggs-usgs:
… appearing in title element.

Before making a pull request
----------------------------

- [x] Make sure all tests run
- [x ] Update the changelog appropriately

WDFN85 -- bug fix
-----------
The name of the monitoring location would no longer appear in the title element after the initial fix for WDFN-85 was applied.

Description
-----------
The code for the adding the monitoring location name to the title was moved outside the 'extra_head_tags' block.  

After making a pull request
---------------------------

### DIFF
--- a/waterdata/templates/monitoring_location.html
+++ b/waterdata/templates/monitoring_location.html
@@ -5,7 +5,6 @@
 {% block extra_head_tags %}
     {% if status_code == 200 %}
         {% if stations|length == 1 %}
-            {% set page_title = stations[0].station_nm %}
             <meta name="description" content="{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}">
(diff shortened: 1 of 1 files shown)


waterdata/templates/monitoring_location.html
```diff
@@ -31,6 +30,8 @@
 
 {% set body_id = 'monitoring-location' %}
 
+{% if stations|length == 1 %}{% set page_title = stations[0].station_nm %}{% endif %}
```
mbucknell: Rather than putting it here, put this line after the import for the macro/components. The reason it didn't work where you put it is that it is only defined within the {% block extra_head_tags %}. By putting it outside of a block it will be available anywhere in the template.

PR #242, opened 2018-04-26
Wdfn 232 - One click to one month, 1 year

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
WDFN 232 - Added radio buttons to change between 7 days, 30 days, and 1 year to show 

Description
-----------
The extended data is only fetched once requested and only for the selected timeseries. Once it is fetched it will be retained in the store. 

I moved some selectors out of components/hydrograph/timeseries and into selector/timeseriesSelectors. Need to figure out when to put a selector in one module versus another. I tried to leave in components/hydrograph/timeseries those selectors that seemed directly related to drawing the hydrograph.

After making a pull request
---------------------------


mbucknell:
For reviewing this, it probably will help to start by looking at the new Actions, retrieveExtendedTimeseries in store/index.js. The extended timeseries are stored with a tsKey as that indicates whether it is current, compare, median (right now we only do current), the period, and the variable. So for instance, 'current:P30D:45807042'

mbucknell:
Also meant to comment that I have completely removed the sr-only tables for the hydrograph. There is a new JIRA to reimplement this.

assets/src/scripts/store/index.js
```diff
@@ -104,6 +105,48 @@ export const Actions = {
             );
         };
     },
+    retrieveExtendedTimeseries(site, period) {
```
danielnaab: Now that we have this one, it's starting to feel like we could merge retrieveTimeseries, retrieveCompareTimeseries, and retrieveExtendedTimeseries together somehow.

assets/src/scripts/store/index.js
```diff
@@ -104,6 +105,48 @@ export const Actions = {
...
+        return function(dispatch, getState) {
+            const state = getState();
+            const parmCd = getCurrentParmCd(state);
+            const tsKey = period === 'P7D' ? 'current' : `current:${period}:${parmCd}`;
```
danielnaab: It'd be nice to keep the key format isolated to one spot - maybe this could be retrieved via a function call, or hasTimeseries() could take the parts in?

assets/src/scripts/store/index.js
danielnaab: Maybe not too much could be merged - it's just starting to feel complex.

assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -52,12 +54,10 @@ export const extendDomain = function (domain, lowerBoundPOW10) {
 };
 
 
-export const getYDomain = function (pointArrays, currentVar) {
+export const getYDomain = function (pointArrays, currentVarParmCd) {
     let yExtent;
     let scaleDomains = [];
 
-    let currentVarParmCd = currentVar && currentVar.variableCode ? currentVar.variableCode.value : null;
-
```
danielnaab: This stuff you did looks good to simplify things!

It also brings to mind a question I've had for awhile... are you aware of if there are any circumstances non-USGS parameter codes overlapping with USGS ones?

assets/src/scripts/components/hydrograph/drawingData.js
```diff
@@ -83,18 +84,19 @@ export const allPointsSelector = createSelector(
  * @param {String} tsKey
  * @return {Object} of keys are tsId, values are Array of point Objects
  */
-export const pointsByTsKeySelector = memoize(tsKey => createSelector(
+export const pointsByTsKeySelector = memoize(tsKey => memoize(period => createSelector(
```
danielnaab: The new memoization library we're using - fast-memoize - should support multiple arguments.

assets/src/scripts/selectors/timeseriesSelector.js
```diff
@@ -0,0 +1,75 @@
...
+    return tsKey === 'median' || period === 'P7D' ? tsKey : `${tsKey}:${period}:${parmCd}`;
+};
+/*
+ * @param {String} tsKey - current, compare, or median
+ * @param {String} or null period - date range of interest specified as an ISO-8601 duration. If null P7D is assumed
+ * @param {String} or null parmCD - Only need to specify if period is something other than P7D or null
+ * @return {Boolean} - True if the time series with key, period, and parmCd has already been requested
+ *
+ */
+export const hasTimeseries = memoize(tsKey => memoize(period => memoize(parmCd => {
```
danielnaab: Same here, I think you can do memoize((tsKey, period, parmCd) => ...)

assets/src/scripts/index.js
```diff
@@ -31,11 +31,13 @@ function main() {
 
     } catch (err) {
         // Send exception to Google Analytics.
-        window.ga('send', 'exception', {
-            // exDescription will be truncated at 150 bytes
-            exDescription: err.stack,
-            exFatal: true
-        });
+        if (window.ga) {
```
danielnaab: 😆 I love it when exception handling code raises its own exceptions!

assets/src/scripts/components/hydrograph/domain.js
jkreft-usgs: parameter codes are a USGS-specific thing.  Other agencies may have parameter codes(EPA used to use them), but there is no authority management.

danielnaab: Yes - so WaterML has that distinction, between variables (globally unique) and variableCodes (organizationally unique). I'm wondering if we have any chance of the database having organizationally unique code overlap? How many outside agency sites does the DB even have?

assets/src/scripts/selectors/timeseriesSelector.js
```diff
@@ -0,0 +1,75 @@
...
+export const getCurrentParmCd = createSelector(
+    getCurrentVariable,
+    (currentVar) => {
+        return currentVar && currentVar.variableCode ? currentVar.variableCode.value : null;
+    }
+);
+
+const tsRequestKey = function(tsKey, period, parmCd) {
+    return tsKey === 'median' || period === 'P7D' ? tsKey : `${tsKey}:${period}:${parmCd}`;
+};
```
danielnaab: I'm a little confused by what's going on here - could we put P7D in the key for consistency?

assets/src/scripts/store/index.js
mbucknell: It's probably worth thinking about. Maybe there are some helper functions that could help.

assets/src/scripts/store/index.js
mbucknell: Agreed. I look at reworking this.

assets/src/scripts/components/hydrograph/drawingData.js
mbucknell: I'll take a look.

assets/src/scripts/index.js
mbucknell: Yeah. I had this happen more than once before this, but with the scope of changes here, I really needed my exception traces!

assets/src/scripts/selectors/timeseriesSelector.js
mbucknell: First question's answer: I don't think so. The issue is that sometimes we use tsKey as one "current", "compare", or "median". The request key is a little different as it's the key used in the requests. These end up being "current", "compare", "median", and then keys with whatever extended data we have retrieved.

second question's answer: That is what I would expect to see in series.timeSeries

danielnaab: Oh the request key, gotcha

assets/src/scripts/components/hydrograph/domain.js
jkreft-usgs: For the purposes of this question, none.  There are sites managed by USGS that are paid for by other agencies- that's generally where the agency code comes from- but they are otherwise indistinguishable from site with USGS as the agency code

assets/src/scripts/selectors/timeseriesSelector.js
mbucknell: So re the first question: I will change the key to use the P7D except for the median. I'm going to wait and see what Andrew comes up with before changing anything about median data.

mbucknell: Actually I'm going to wait on this. It is turning into a bigger deal and I'd like to get this code out.

ayan-usgs: So, I'm still working on the rejiggering the median data, here's what I I've been scheming so far: https://github.com/ayan-usgs/waterdataui/blob/wdfn-221-median-data-step-function/assets/src/scripts/components/hydrograph/statistics.js#L10

That function is intended to be able to figure out what dates to display based on the date range of the timeseries. Hopefully, that'll allow work to continue while I get the median step function sorted out.

PR #245, opened 2018-04-27
Wdfn 232 - part2

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run
- [ ] Update the changelog appropriately - no changes needed

Title
-----------
WDFN 232 - Cleaned up X axis display and made the tsRequestKey formation more consistent

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/axes.js
```diff
@@ -10,12 +10,33 @@ const { layoutSelector } = require('./layout');
...
+};
+
+const tickInterval = function(period) {
+    switch (period) {
+        case 'P7D':
+            return timeDay;
+        case 'P30D':
+            return timeWeek;
+        case 'P1Y':
+            if (mediaQuery(USWDS_LARGE_SCREEN)) {
```
danielnaab: As an aside, looking into the image server I realized we'll have a challenge with these JS media queries, especially when SVG is the output format as it doesn't have a fixed size. Something to think about...

assets/src/scripts/selectors/timeSeriesSelector.js
```diff
@@ -39,37 +39,40 @@ export const getCurrentParmCd = createSelector(
...
+ * @param {String} or null period - date range of interest specified as an ISO-8601 duration. If null, P7D is assumed
+ * @param {String} or null parmCD - Only need to specify if period is something other than P7D or null
+ * @return {Boolean} - True if the time series with key, period, and parmCd has already been requested
+ *
+ */
+export const hasTimeSeries = memoize((tsKey, period, parmCd) => createSelector(
+    getTsRequestKey(tsKey, period, parmCd),
+    state => state.series,
+    (tsRequestKey, series) => {
+        return series && series.requests && series.requests[tsRequestKey] ? true : false;
```
danielnaab: Here you could omit the `? true : false` part

mbucknell: True, but then the result will be "truthy" or "falsy" not true or false. I may be being to strict about this but the comment says its returning a boolean so I'm returning a Boolean

danielnaab: Ah, okay. Maybe `Boolean(...)`?

mbucknell: 👍

assets/src/scripts/store/index.js
```diff
@@ -87,31 +89,32 @@ export const Actions = {
...
+                () => dispatch(Actions.resetTimeseries(getTsRequestKey('compare', 'P7D')(getState())))
             );
         };
     },
     retrieveExtendedTimeseries(site, period) {
         return function(dispatch, getState) {
             const state = getState();
             const parmCd = getCurrentParmCd(state);
-            const tsKey = period === 'P7D' ? 'current' : `current:${period}:${parmCd}`;
+            const requestKey = getTsRequestKey ('current', period, parmCd)(state);
```
danielnaab: This seems like it should work well

assets/src/scripts/selectors/timeSeriesSelector.spec.js
```diff
@@ -152,76 +152,130 @@ describe('timeSeriesSelector', () => {
...
             series: {
+                variables: TEST_VARS,
                 requests : {
-                    'current': {},
+                    'current:P7D': {},
                     'median' : {},
                     'current:P30D:00060': {}
                 }
+            },
+            timeseriesState: {
```
ayan-usgs: Should this be changed to `timeSeriesState` in the tests and in the store? Not sure if we want this consistency to extend into the store as well. :thinking:

PR #255, opened 2018-05-01
Wdfn 251

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Show estimated more clearly
-----------

Description
-----------
The Estimated qualifier uses a lowercase 'e'. We were assuming uppercase. Styling estimated with a solid red line.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/drawingData.js
+++ b/assets/src/scripts/components/hydrograph/drawingData.js
@@ -171,7 +171,7 @@ export const flatPointsSelector = memoize(tsKey => createSelector(
 export const classesForPoint = point => {
     return {
         approved: point.qualifiers.indexOf('A') > -1,
-        estimated: point.qualifiers.indexOf('e') > -1
+        estimated: point.qualifiers.indexOf('e') > -1 || point.qualifiers.indexOf('E') > -1
     };
 };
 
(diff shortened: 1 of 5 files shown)


assets/src/scripts/components/hydrograph/drawingData.js
```diff
@@ -171,7 +171,7 @@ export const flatPointsSelector = memoize(tsKey => createSelector(
 export const classesForPoint = point => {
     return {
         approved: point.qualifiers.indexOf('A') > -1,
-        estimated: point.qualifiers.indexOf('E') > -1
+        estimated: point.qualifiers.indexOf('e') > -1
```
danielnaab: Based on Janell's comments here: https://internal.cida.usgs.gov/jira/browse/WDFN-71?focusedCommentId=139138&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-139138

... "E" and "e" mean different things:

```
'E' => ['0', 'e', 'Value was computed from estimated unit values.'],
'e' => ['0', 'e', 'Value has been estimated.'], # Aquarius ESTIMATE
```

Wondering if we need to support both?

mbucknell: And do we every see a 'E'? I don't really understand the 'E' description above. This seems like a @bdgarner-usgs question.

mbucknell: Also found this: https://help.waterdata.usgs.gov/codes-and-parameters/instantaneous-value-qualification-code-uv_rmk_cd. 

That makes me inclined to treat both 'E' and 'e' qualifier codes as estimated.

PR #259, opened 2018-05-02
Wdfn 221 median data step function

ayan-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -182,42 +183,23 @@ const timeSeriesLegend = function(elem) {
...
-    }
+const plotMedianPoints = function (elem, {xscale, yscale, modulo, points}) {
+    let stepFunction = d3Line()
+        .curve(curveStepAfter)
+        .x(function(d) {
+            return xscale(d.dateTime);
+        })
+        .y(function(d) {
+            return yscale(d.value);
+        });
```
danielnaab: That's pretty nice

assets/src/scripts/components/hydrograph/statistics.js
```diff
@@ -0,0 +1,58 @@
...
+        filtered.unshift(leftVal);
+    }
+    let last = filtered[filtered.length - 1];
+    if (last.dateTime < series.endTime) {
+        let rightVal = Object.assign({}, last);
+        rightVal.dateTime = series.endTime;
+        filtered.push(rightVal);
+    }
+    return filtered;
+};
```
danielnaab: This function looks good, but could you add in a comment explaining each section of it?

danielnaab:
I'm thinking that if we calculated time series-like data rather than store on the state, we should put the data outside of the `series` section of the state - which could be reserved for use of data from the IV service. Would that make sense? I'm not sure if that would cause some mixups on key references anywhere...

assets/src/scripts/components/hydrograph/statistics.js
danielnaab: I need to run - will finish looking at this tomorrow.

ayan-usgs:
> I'm thinking that if we calculated time series-like data rather than store on the state, we should put the data outside of the series section of the state - which could be reserved for use of data from the IV service.

I've been thinking that too. I would take on that task a separate ticket though because that refactor would likely ripple into other parts of the code base.

mbucknell:
I agree with the discussion regarding moving the median data out of the series state. Another ticket sounds like the right thing to do.

PR #270, opened 2018-05-04
Wdfn234 add sponsorship logos

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-234 Add sponsorship data from SIFTA
-----------
Added call to a web service providing cooperator logos, names, and links. These are now displayed at the bottom of the page above the footer.

Description
-----------
WDFN-234 Add sponsorship data from SIFTA
Note that the current web call accesses a temporary service, code will need to change when the permanent service becomes available.

After making a pull request
---------------------------


wdfn-server/waterdata/tests/test_views.py
```diff
@@ -48,6 +49,13 @@ def test_everything_okay(self, r_mock):
         m_resp_param.iter_lines.return_value = iter(self.parameter_lines)
         r_mock.side_effect = [m_resp, m_resp_param]
 
+        m_resp_lookup = mock.Mock()
+        m_resp_lookup.status_code = 200
+        m_resp_lookup.return_value = {'Customers': [{'Name': 'mock_cooperator',
+                                                     'URL': 'http://www.fake.gov',
+                                                     'IconURL': 'http://www.fake.gov'}]}
+        r_mock_lookup.return_value.json = m_resp_lookup
```
mbucknell: This doesn't make sense. . The return value for r_mock_lookup is just a dictionary not an Object where you would need to define the json function. I think you can eliminate lines 52, 53 and then combine 54 and 57 to assign r_mock_lookup.return_value to the dictionary.

wdfn-server/waterdata/tests/test_views.py
```diff
@@ -67,6 +75,8 @@ def test_everything_okay(self, r_mock):
         )
         self.assertEqual(json_ld_response.status_code, 200)
         self.assertIsInstance(json.loads(json_ld_response.data), dict)
+        # see if call to Cooperator Lookup Service returns expected response
+        self.assertEqual(m_resp_lookup.status_code, 200)
```
mbucknell: This is unnecessary and is not actually testing code in the view.  Your asserts should be looking at the response object. The only thing you might be able to do is check the rendered template to see if the sifta data made it there. That's overkill and would be brittle so just eliminate lines 78 and 79.

wdfn-server/waterdata/utils.py
```diff
@@ -92,3 +92,26 @@ def wrap(f):
         return update_wrapper(func, f)
 
     return wrap
+
+
+def execute_lookup_request(url_root_cooperator_lookup, url_path_cooperator_lookup, params):
```
mbucknell: This function is specific to the cooperator lookup service so consider renaming execute_lookup_request to something like execute_cooperator_lookup_request.

abriggs-usgs: All good advice. Much Appreciated.

PR #279, opened 2018-05-07
WDFN-243 - SVG graph service

danielnaab:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-243
-----------
SVG graph service

Description
-----------
This PR adds Express, and is configured to serve up by default at `localhost:2929`. An example URL for the SVG service is at `http://localhost:2929/monitoring-location/07048495/?parameterCode=00065`. There are two headless browser renderers - Puppeteer (works), PhantomJS (doesn't work). I'm not certain if we should use either when all is said and done, but I found having two implementations helped hash out the code organization a bit.

Note that this does not do concurrent requests well - the focus was mainly on getting the rendering code to return something sensible.

After making a pull request
---------------------------


graph-server/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+    renderToRespone(res, {
+        renderer: req.query.renderer,
+        siteID: req.params.siteID,
+        parameterCode: req.query.parameterCode,
+        compare: req.query.compare
+    });
+});
+
+
+module.exports = server;
```
mbucknell: Do we have any rules around when we use module.exports vs export const nameOfFunction? Question really for anyone.

graph-server/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+});
+
+
+app.get('/monitoring-location/:siteID/', checkSchema({
+    renderer: {
+    },
+    parameterCode: {
+        in: ['query'],
+        errorMessage: 'parameterCode (5 digit integer) is required',
+        isInt: true,
```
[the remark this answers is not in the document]
danielnaab: Yes - this will verify that the string corresponds to an integer, but not coerce it to an integer object (the `toInt` sanitizer would be used for that).

graph-server/src/index.js
danielnaab: In the case of node.js, we can only use the CommonJS module.exports style. Don't know when ES6-style will officially land in node. It wouldn't hurt to be more consistent...

graph-server/src/renderer/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+const getSvgImpl = {
+    puppeteer: require('./impl/puppeteer'),
+    phantomjs: require('./impl/phantomjs')
+};
+const DEFAULT_IMPLEMENTATION = 'puppeteer';
+
+
+const SERVICE_ROOT = 'https://waterservices.usgs.gov/nwis';
+const PAST_SERVICE_ROOT = 'https://nwis.waterservices.usgs.gov/nwis';
```
mbucknell: Eventually this should come from config. We could potentially use local urls for this?

danielnaab: Agree... on the watermark/legend branch I started grabbing them from environment variables. Don't know if that is best, or a config file would be better.

danielnaab:
Maybe we could iterate over all the implementations, and run the entire test suite on each.. Assuming we keep multiple renderers.

PR #287, opened 2018-05-09
Wdfn 277 - Safari hangs when drawing a new graph that has a median line

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
xscale no longer returns a scale with the domain set to [0 ,1]

Description
-----------
Also reverted the changes that were made to set the currentDateRange after the data was fetched.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/scales.js
+++ b/assets/src/scripts/components/hydrograph/scales.js
@@ -18,12 +18,15 @@ const { getVariables, getCurrentParmCd, getRequestTimeRange } = require('../../s
  * @return {Object} d3 scale for time.
  */
 function createXScale(timeRange, xSize) {
-    const xExtent = timeRange ? [timeRange.start, timeRange.end] : [0, 1];
+    //const xExtent = timeRange ? [timeRange.start, timeRange.end] : [0, 1];
 
+    let scale = scaleTime()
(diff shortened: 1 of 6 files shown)


assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -18,12 +18,13 @@ const { getVariables, getCurrentParmCd, getRequestTimeRange } = require('../../s
...
-
     // xScale is oriented on the left
-    return scaleTime()
-        .range([0, xSize])
-        .domain(xExtent);
+    let scale = scaleTime()
+        .range([0, xSize]);
+    if (timeRange) {
+        scale.domain([timeRange.start, timeRange.end]);
+    }
```
danielnaab: Curious - what does `scale.domain()` return here, if `timeRange` isn't set? I thought `[0, 1]` is always the default with D3. Must be reading the code wrong... but scaleTime is a continuous scale: https://github.com/d3/d3-scale/blob/master/src/continuous.js#L7

mbucknell: default is [2000-01-02, 200-01-02]. see https://github.com/d3/d3-scale/blob/master/README.md#scaleTime

PR #294, opened 2018-05-11
Wdfn 226 reflect local time

ayan-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/timeSeries.js
```diff
@@ -136,3 +137,18 @@ export const descriptionSelector = createSelector(
...
+    getIanaTimeZone,
+    getNwisTimeZone,
+    (ianaTimeZone, nwisTimeZone) => {
+        const tz = DateTime.fromObject({zone: ianaTimeZone});
+        const tzShortName = tz.offsetNameShort.toUpperCase();
+        const nwisShortNames = Object.values(nwisTimeZone).map(tz => tz.zoneAbbreviation.toUpperCase());
+        if (nwisShortNames.includes(tzShortName)) {
+            return ianaTimeZone;
+        } else {
+            return 'UTC';
```
ayan-usgs: I'm on the fence about whether using UTC as backup timezone would be confusing... :thinking:

mbucknell: I do think it should return local time zone if ianaTimeZone is null.

assets/src/scripts/store/index.js
```diff
@@ -47,19 +47,28 @@ const getCurrentVariableId = function(timeSeries, variables) {
...
             dispatch(Actions.addTimeSeriesLoading([requestKey]));
-
             const timeSeries = getTimeSeries({sites: [siteno], params}).then(
                 series => {
                     const collection = normalize(series, requestKey);
 
                     // Get the start/end times of this request's range.
                     const notes = collection.queryInfo[requestKey].notes;
-                    const endTime = notes.requestDT;
+                    const endTime = new Date(notes.requestDT);
```
mbucknell: I wonder if we can figure out startTime without converting to a Date object by subtracting the appropriate number of milliseconds.

mbucknell: Or maybe use calcStartTime. We know that we have requested P7D so getting the periodDays from the filter is really not necessary.

assets/src/scripts/store/index.js
```diff
@@ -294,6 +302,12 @@ export const Actions = {
             type: 'SET_GAGE_HEIGHT',
             gageHeight
         };
+    },
+    setLocationIANATimeZone(ianaTimeZone) {
```
mbucknell: I've been starting to follow best practices as described here: https://medium.com/@kylpo/redux-best-practices-eef55a20cc72. They recommend <NOUN><VERB> for action constants NOUN_VERB
.

assets/src/scripts/store/index.spec.js
```diff
@@ -55,6 +59,42 @@ describe('Redux store', () => {
             }
         };
 
+        describe('retrieveLocationTimeZone with good data', () => {
```
mbucknell: It would be great to add a test for bad data too.

assets/src/scripts/store/index.spec.js
```diff
@@ -836,6 +876,8 @@ describe('Redux store', () => {
     });
 });
 
+const MOCK_WEATHER_SERVICE_DATA = '{"properties" : {"timeZone" : "America/Chicago"}}';
```
mbucknell: I think it would be fine for this to go in the actual test suite for retrieveLocationTimeZone. It's not long and I think it should be the exception to separate test data from the test suite.

assets/src/scripts/store/seriesReducer.js
```diff
@@ -20,6 +20,13 @@ const resetTimeSeries = function(series, action) {
     };
 };
 
+const addIanaTimeZone = function(series, action) {
```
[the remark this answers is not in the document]
danielnaab: @mbucknell That's a good point, and we could get the site number off state without adding anything new - it's in `sourceInfo` / `siteCode`.

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -1,5 +1,5 @@
-{% macro TimeSeriesComponent(site_no) -%}
-    <div class="wdfn-component" data-component="hydrograph" data-siteno="{{ site_no }}">
+{% macro TimeSeriesComponent(site_no, latitude, longitude) -%}
+    <div class="wdfn-component" data-component="hydrograph" data-siteno="{{ site_no }}" data-latitude="{{ latitude }}" data-longitude="{{ longitude }}">
         <div class="loading-indicator-container"></div>
```
danielnaab: This is going to require some changes to the image server to pass this in, also.. but could be done as a separate ticket.

assets/src/scripts/components/hydrograph/timeSeries.js
ayan-usgs: The only site that I've found so far is 251549080251200, which the weather service reports as being in UTC, but NWISWeb reports as being in eastern time.

assets/src/scripts/store/seriesReducer.js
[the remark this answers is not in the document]
ayan-usgs: I cool with writing up a separate ticket for this.

wdfn-server/waterdata/templates/macros/components.html
ayan-usgs: I'll write a separate ticket for this.

assets/src/scripts/components/hydrograph/timeSeries.js
mbucknell: The reason I thought local would be better is because NWIS will typically report two timezones (ST and DT). Which to use?

ayan-usgs: So from the NWIS time zones (e.g. EST, MST, etc) and NWIS code about whether or not a site observes daylight savings, I think we could potentially infer the correct IANA code. Then we could let luxon figure out when to use ST or DT.

My main concern would be the reliability of inferring the timezone. It seems more predictable to fall back to local rather than to attempt the inference.

ayan-usgs: Mainly thinking about situations such as Arizona or those parts of Indiana (prior to 2006) that don't observe DT. Things might get messy real quick if we attempt the inference.

danielnaab: Yeah that makes sense

assets/src/scripts/store/index.js
ayan-usgs: Opted to use `calcStartTime`.

PR #300, opened 2018-05-16
Wdfn 263 address aria violations

ayan-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


mbucknell:
I'm inclined to leave the h3 tag in the provisional alert as is. This comes from the USWDS standards. We could potentially use the h2 tag and then style appropriately. Did you test to see what it looks like if you just use a standard <div>?

danielnaab:
It's possible to do `@include h3()` in sass. Curious why that's a violation?

ayan-usgs:
> I'm inclined to leave the h3 tag in the provisional alert as is. This comes from the USWDS standards. We could potentially use the h2 tag and then style appropriately. Did you test to see what it looks like if you just use a standard div tag ?

I'm inclined to leave it as it is as well. It's definitely a bit more subdued as a div with bolded text.

ayan-usgs:
> It's possible to do @include h3() in sass. Curious why that's a violation?

I think it's because we have an h1 tag for the station name... then the next header is an h3 without there having been an h2.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -236,6 +236,7 @@ const createTitle = function(elem) {
 const watermark = function (elem) {
     elem.append('img')
         .classed('watermark', true)
+        .attr('alt', 'USGS logo')
```
danielnaab: This is a nit, but I think this should represent what's on the logo.. perhaps: "USGS - science for a changing world".

ayan-usgs: I'm cool with that.

PR #301, opened 2018-05-16
Wdfn 261

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Refactor median data out of the time series section of the application's redux store

Description
-----------
The median statistics are now in the statisticsData property. This will be an object, with the keys being a statistic name. Right now we only have median. I have only created a medianSelector module as a result. But this should allow us to more easily add other statistics to the store. 

The generation of the median points to be fed to the plotting call is in drawingData.

After making a pull request
---------------------------


mbucknell:
I have already found one thing that I did not do and that is change the yScale to include the median data.  So there will be changes.

PR #303, opened 2018-05-17
SauceLabs

danielnaab:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Description
-----------
This PR adds a Sauce Labs test launcher to the Karma configuration, and updates TravisCI to run them in a parallel build. This is done to speed up the job, and also allow us to ignore the success/failure status - the results drive the browser matrix that is displayed on the README.

After making a pull request
---------------------------


assets/karma.conf.js
```diff
@@ -25,20 +28,19 @@ module.exports = function (config) {
...
-
         // preprocess matching files before serving them to the browser
         // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
         preprocessors: {
             'src/scripts/**/*.js': ['browserify']
         },
+
         browserify: {
-            debug: true,
+            debug: process.argv.some(isDebug),
```
mbucknell: So how do we run with debug now? I wonder if we can add it to the scripts, at least test:watch.

danielnaab: I didn't intend to change behavior, but to condense the logic in this section: https://github.com/usgs/waterdataui/blob/master/assets/karma.conf.js#L299-L309

... which was I believe overwriting the debug value set here.

PR #306, opened 2018-05-17
Wdfn 283 location metadata

ayan-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -486,6 +486,8 @@ const attachToNode = function (store, node, {siteno, parameter, compare, cursorO
     };
     const latitude = node.dataset.latitude;
     const longitude = node.dataset.longitude;
+    const locationNumber = node.dataset.siteno;
```
mbucknell: Some things to think about. I think latitude and longitude are also metadata. Consider setting the locationIdentifier and latitude/longitude in scripts/index.js when the store is configured. The map component also uses siteno and lat/long. In both the hydrograph and map components you can use the necessary selector directly with the passed in store object to retrieve the identifier or the lat/long. The expectation for these is they wouldn't change so I don't think they need Actions to set them.

ayan-usgs: Sounds good.

PR #309, opened 2018-05-21
adding change to fix the background opacity of the tooltipTextGroup

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-287
-----------
Fixed a problem where tooltips which overlapped graph lines were hard to read

Description
-----------
Made change to opacity of tooltip group background

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/tooltip.js
```diff
@@ -137,7 +137,7 @@ const createTooltipTextGroup = function (elem, {currentPoints, comparePoints, qu
         const baseMarginOffsetTextGroup = 27;
         let marginAdjustment = layout.margin.left + baseMarginOffsetTextGroup;
         elem.style('margin-left', marginAdjustment + 'px');
-
+        elem.style('background', 'rgba(255, 255, 255, 0.8');
```
danielnaab: Could you put this as a style in app.scss instead? Looks like rule nested a level under `.tooltip-text-group`?

abriggs-usgs: Maybe. I tried it there first, but it seemed that using '.style' to format the margin, and having the background styling in 'app.scss,' caused the background styling to be applied to the entire element including the margin-which was weird since the margin is normally transparent. Anyway, doing it that way caused the y-axis to be obscured by the background opacity; not totally sure why.

danielnaab: Could it be it was set on `.tooltip-text-group` rather than `.tooltip-text-group div`?

abriggs-usgs: Oh, I think I see what you mean now . . .

PR #313, opened 2018-06-01
Switch to Rollup and Bublé

danielnaab:
If you want to run this locally, you'll need to manually edit Leaflet's `node_modules/leaflet/package.json`, adding `"module": "src/Leaflet.js"`.

Before making a pull request
----------------------------

- [ ] Make sure all tests run

Description
-----------
This removes Babel and Browserify in favor of Bublé and Rollup, with ES6 modules. babel-polyfill was replaced with explicit imports of the polyfills we need. The advantages are:

- Easier ability to code-split bundles and do asynchronous loading
- Bundle size reduced from 680kb to 450kb
- Faster compilation with Bublé

Some things to note:
- The `karma-rollup-preprocessor` used for running the tests compiles each spec file as a separate bundle - which is slow and resource-intensive. So an entry point was created that imports all the specs into a single bundle, making the tests run quickly.
- Initially (when working with Webpack) I replaced proxyquire with a Webpack loader, inject-loader. But Rollup doesn't have a good solution, so I replaced the mocks with dependency injection. I think this might be a better general solution, so the code is not so bound to the transpiler we use. However, a couple of them feel cumbersome.

After making a pull request
---------------------------


assets/rollup.config.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * Rollup configuration.
+ * NOTE: This is a CommonJS module so it can be imported by Karma.
+ */
+
+const buble = require('rollup-plugin-buble');
```
ayan-usgs: Not a Javascript expert, so I'm going ask the question: why can't you use ES6's import syntax here?

danielnaab: I wanted to import from the Karma config... since that's run under Node this one needed to be CommonJS also.

PR #314, opened 2018-06-05
Sifta json

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-292 - SIFTA Logos, Change to Use JSON file
-----------
Changed cooperator logos lookup to use json file

Description 
-----------
After it was noted that the SIFTA site service only worked when accessed from an internal network, the code was reconfigured to use a JSON file containing the SIFTA (cooperator information). The current JSON file only contains information for districts 20 and 51 (Kansas and Virginia).

After making a pull request
---------------------------


wdfn-server/waterdata/views.py
```diff
@@ -170,6 +161,26 @@ def return_404(*args, **kwargs):
     return abort(404)
 
 
+def get_cooperator_data(location_with_values, site_no):
```
mbucknell: I would prefer that this go into utils. We should only

wdfn-server/waterdata/views.py
```diff
@@ -170,6 +161,26 @@ def return_404(*args, **kwargs):
...
+    """
+    Gets the cooperator data from a json file, currently a feature toggle, and limited to district codes 20 and 51
+
+    :param site_no: USGS site number
+    :param location_with_values: monitoring location data including district code
```
mbucknell: Could we just pass in district_cd, rather than the whole location?

wdfn-server/waterdata/views.py
abriggs-usgs: Good idea

wdfn-server/waterdata/views.py
abriggs-usgs: Yep. And will do.

PR #377, opened 2018-11-05
Wdfn 302 ticks below xaxis fix

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-302 Part 2, Bug fix for tick marks below x-axis
-----------
Added logic to stop the additional tick marks from showing below the x-axis
Fixed the uneven display of negative vs positive tick marks
Made rounding of numbers relative to the size of the tick value rounded 

Description
-----------
Repairs several issues found while testing

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/domain.js
+++ b/assets/src/scripts/components/hydrograph/domain.js
@@ -156,7 +156,7 @@ export const getRoundedTickValues = function(additionalTickValues, yDomain) {
         roundedTickValues.push(value);
         }
     });
-console.log("this is it " + JSON.stringify(roundedTickValues))
+
     return roundedTickValues;
 };
 
--- a/assets/src/scripts/components/hydrograph/domain.js
+++ b/assets/src/scripts/components/hydrograph/domain.js
@@ -7,7 +7,6 @@ import { visiblePointsSelector } from './drawingData';
 import { getCurrentParmCd } from '../../selectors/timeSeriesSelector';
 
 
-const MULTIPLE_TO_ROUND_TO = 5;
 const PADDING_RATIO = 0.2;
 const Y_TICK_COUNT = 5;
 // array of parameters that should use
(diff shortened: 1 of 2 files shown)


assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -112,29 +111,108 @@ export const getLowestAbsoluteValueOfTickValues = function(tickValues) {
...
+ * @returns {array} roundedTickValues, numerical values for y-axis ticks rounded to a multiple of a given number
+ */
+export const getRoundedTickValues = function(additionalTickValues) {
+    let roundedTickValues = [];
+    // round the values based on an arbitrary breakpoints and rounding targets (may result in duplicate array values)
+    additionalTickValues.forEach(function(value) {
+        if (value > 1000) {
+            value = Math.ceil(value/1000)*1000;
+            roundedTickValues.push(value);
+        } if (value > 100) {
```
danielnaab: Did you mean this to be `else if`?

assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -112,29 +111,108 @@ export const getLowestAbsoluteValueOfTickValues = function(tickValues) {
...
+ * Helper function that rounds numerical values in an array based on the numerical value of the individual array item
+ * and a somewhat arbitrary numeric value, a multiple of which the targeted array item's value is rounded.
+ * @param {array} additionalTickValues, numerical values for y-axis ticks
+ * @returns {array} roundedTickValues, numerical values for y-axis ticks rounded to a multiple of a given number
+ */
+export const getRoundedTickValues = function(additionalTickValues) {
+    let roundedTickValues = [];
+    // round the values based on an arbitrary breakpoints and rounding targets (may result in duplicate array values)
+    additionalTickValues.forEach(function(value) {
+        if (value > 1000) {
```
danielnaab: Wondering if the problem this fixes would also happen with `values > 10000`?

assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -112,29 +111,108 @@ export const getLowestAbsoluteValueOfTickValues = function(tickValues) {
...
+    let lowestTickValueOfLogScale = getLowestAbsoluteValueOfTickValues(tickValues);
+
+    // Make a set of tick values that when graphed will have equal spacing on the y axis
+    additionalTickValues = generateAdditionalTickValues(lowestTickValueOfLogScale);
+
+    // round the values to a chosen multiple of a number
+    additionalTickValues = getRoundedTickValues(additionalTickValues);
+
+    // remove any tick values that would be below the y-axis
+    additionalTickValues = removeTickValuesBelowXaxis(additionalTickValues, yDomain);
```
danielnaab: Could you instead pass `yDomain` into `getRoundedTickValues`, and just skip over values outside the domain (so `removeTickValuesBelowXaxis` becomes unnecessary)?

assets/src/scripts/components/hydrograph/domain.js
abriggs-usgs: Yuss, I did.

assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -112,31 +111,96 @@ export const getLowestAbsoluteValueOfTickValues = function(tickValues) {
...
+ */
+export const getRoundedTickValues = function(additionalTickValues, yDomain) {
+    let roundedTickValues = [];
+    // round the values based on an arbitrary breakpoints and rounding targets (may result in duplicate array values)
+    additionalTickValues.forEach(function(value) {
+        let roundingFactor = 1;
+        // first check that the value is greater than the position where the x-axis intersects the y-axis, then round
+        if (value > yDomain[0]) {
+            if (value > 10000) {
+               roundingFactor = 10000;
```
danielnaab: I was wondering if maybe a more general solution might be possible, using Math.log10 / Math.pow. I think this is fine as-is, though.

danielnaab:
Looks good 👍

assets/src/scripts/components/hydrograph/domain.js
abriggs-usgs: Yeah, I don't like the clunky look of the code here, but I did mess with a couple 'one-size-fits-all' attempts and I couldn't find anything that looked good across the gamut of data sets. :(

PR #422, opened 2019-02-20
Wdfn 317 - Page dies due to zero values

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Page dies due to zero values for all times in the time series

Description
-----------

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/domain.js
+++ b/assets/src/scripts/components/hydrograph/domain.js
@@ -68,7 +68,11 @@ export const getYDomain = function (pointArrays, currentVarParmCd) {
             // this isn't useful for creation of data
             // points, so add this broadens the extent
             // a bit for single point series
-            ptExtent = [ptExtent[0] - ptExtent[0]/2, ptExtent[0] + ptExtent[0]/2];
+            if (ptExtent[0]) {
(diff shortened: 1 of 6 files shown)


assets/src/scripts/components/hydrograph/domain.js
```diff
@@ -68,7 +68,11 @@ export const getYDomain = function (pointArrays, currentVarParmCd) {
             // this isn't useful for creation of data
             // points, so add this broadens the extent
             // a bit for single point series
-            ptExtent = [ptExtent[0] - ptExtent[0]/2, ptExtent[0] + ptExtent[0]/2];
+            if (ptExtent[0]) {
+                ptExtent = [ptExtent[0] - ptExtent[0] / 2, ptExtent[0] + ptExtent[0] / 2];
```
abriggs-usgs: Just attempting to understand what is happening here. Won't ptExtent[0] - ptExtent[0]/2 always be zero? And won't ptExtent[0] + ptExtent[0] / 2 always be ptExtent[0]?

mbucknell: The else will be taken if ptExtent[0] = 0 since that is a falsy value in javascript (see https://developer.mozilla.org/en-US/docs/Glossary/Falsy). That's exactly what was happening before I put the if statement in since there was no guard.

PR #434, opened 2019-02-27
Graph server Dockerization and hardening

danielnaab:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Description
-----------
This includes Docker containers that will (hopefully) work in a production deployment, and changes to how the Puppeteer-based PNG screenshots are generated. A process pool with a configurable pool size is used to increase the number of concurrent images that may be generated.

After making a pull request
---------------------------


docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+# This base config defines the production containers.
+
+version: "3"
```
mbucknell: Our pattern has been to use a specific version in our compose files.

docker-compose.override.yml
```diff
@@ -0,0 +1,24 @@
...
+# Put custom config in local.env.
+
+version: "3"
+
+services:
+  assets:
+    command: npm run watch
+    volumes:
+      - ./assets:/assets
+      - ./volumes/assets/node_modules:/assets/node_modules
```
mbucknell: I'm still learning docker. I don't understand what you are mounting here and why?

docker-compose.override.yml
```diff
@@ -0,0 +1,24 @@
...
+  graph_server:
+    volumes:
+      - ./graph-server/src:/graph-server/src
+      - ./assets/dist:/assets
+      - ./volumes/graph-server/node_modules:/graph-server/node_modules
```
mbucknell: See above question.

graph-server/Dockerfile
```diff
@@ -0,0 +1,44 @@
+# Build the static assets in the node.js container.
+FROM wdfn_assets as assets
+RUN npm run build
+
+# Use LTS release
+FROM node:10.15-alpine
+
+# Installs latest Chromium package.
+RUN apk update && apk upgrade && \
+    echo @edge http://nl.alpinelinux.org/alpine/edge/community >> /etc/apk/repositories && \
```
ayan-usgs: ~I see we're pinning to Alpine's edge repository. Should we be concerned about any of these packages changing underneath our feet without us knowing about it?~

ayan-usgs: Nevermind. You want to latest package.

graph-server/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# Add user so we don't need the --no-sandbox Chrome option
+RUN addgroup -S grapher && adduser -S -g grapher grapher \
+    && mkdir -p /home/grapher/Downloads \
+    && chown -R grapher:grapher /home/grapher \
+    && chown -R grapher:grapher /graph-server
+USER grapher
+
+EXPOSE 2929
+
+CMD npm start
```
ayan-usgs: Is there any sort of Docker health check that would be appropriate to add to this?

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    image: wdfn_assets
+    ports:
+      - "9000:9000"
+  # Node server-renderer
+  graph_server:
+    privileged: true
+    build: ./graph-server
+    image: wdfn_graph_server
+    depends_on:
+      - assets
```
ayan-usgs: :+1:

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  # Node-based static asset build tooling
+  assets:
+    build: ./assets
+    image: wdfn_assets
+    ports:
+      - "9000:9000"
+  # Node server-renderer
+  graph_server:
+    privileged: true
```
ayan-usgs: Why is this necessary?

docker-compose.override.yml
danielnaab: I did this so I can have a `node_modules` directory on my macOS host that is different than the one in the container. Otherwise they'll conflict with each other, meaning you need to reinstall them over and over again if developing against both.

graph-server/Dockerfile
danielnaab: Hmm, I'm thinking that could be a separate feature... add an additional route to the server that can be pinged.

docker-compose.yml
danielnaab: I suspect this may be a deployment issue and in general a problem, but Chrome won't run without it. There's follow-up work needed to determine if there is a viable alternative.

PR #446, opened 2019-03-13
Graph server deploy

ayan-usgs:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


Jenkinsfile.deploy
```diff
@@ -0,0 +1,16 @@
+@Library(value='PipelineLibs@0.6.1', changelog=false) _
```
ssoper-usgs: Might want to use `0.6.3` here to mirror the library version in the build file.

graph-server/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# Add manifest and install dependencies
+COPY package.json package-lock.json ./
+RUN npm ci
+
+# Add the server source code
+COPY ./src ./src
+
+# Add customer entrypoing that will help us set environment variables in ECS
+COPY ./bin/docker-entrypoint.sh /usr/local/bin/
+ENTRYPOINT ["docker-entrypoint.sh"]
```
ssoper-usgs: Might be different for node builds, but do we still want to use `entrypoint` here?

Jenkinsfile.deploy
ayan-usgs: > Might want to use `0.6.3` here to mirror the library version in the build file.

I'm actually thinking it's going to be the next version (e.g. `0.7.0`) that we'll be using, so I left as Dan had it. The next library version has a new parameter that's important for the server to work.

graph-server/Dockerfile
ayan-usgs: As of right now, I think so.

PR #625, opened 2019-08-30
Iow 41 - Support running the graph server off a path on the ALB

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Support running the graph server off a path on the ALB

After making a pull request
---------------------------

### DIFF
--- a/graph-server/src/index.js
+++ b/graph-server/src/index.js
@@ -10,7 +10,7 @@ const renderToRespone = require('./renderer');
 
 const PORT = process.env.NODE_PORT || 2929;
 const CACHE_TIMEOUT = 15 * 60;  // 15 minutes
-const PATH_CONTEXT = process.env.NODE_PORT || '/api/graph-images';
+const PATH_CONTEXT = process.env.PATH_CONTEXT || '/api/graph-images';
 
 
 // Create the Express app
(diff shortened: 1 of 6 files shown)


graph-server/src/index.js
```diff
@@ -10,6 +10,7 @@ const renderToRespone = require('./renderer');
 
 const PORT = process.env.NODE_PORT || 2929;
 const CACHE_TIMEOUT = 15 * 60;  // 15 minutes
+const PATH_CONTEXT = process.env.NODE_PORT || '/api/graph-images';
```
ayan-usgs: So, I could either use the default or I could pass in a `NODE_PORT` environment variable to set the context. Am I understanding that right?

mbucknell: Potentially but at the moment I don't think either the docker file or the npm script has been set up to use that. That's also a typo that I have fixed. Should be PATH_CONTEXT. But regardless, I think some changes would need to be made in Dockerfile and package.json to use the environment variables.

ayan-usgs: >I think some changes would need to be made in Dockerfile and package.json to use the environment variables.

Yep. Thought we were using one our images, but we're actually using a clean version of Alpine off Docker Hub.

ayan-usgs: Suffice it to say, I think this achieves what we set out to do.

PR #637, opened 2019-09-13
IoW-37 Camera Images on WDFN site pages

duselmann:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------
IoW-37 Added Monitoring Camera images to WDFN

Description
-----------
This is a draft implementation to show that we can pull display the camera images.
Future improvements should include a means to update the camera json file - likely a service of some sort. We should also consider liking the images with the UI. Since only the 99 most recent images are preserved on the image host, if we want more then consider a means of preserving them ourselves.

After making a pull request
---------------------------


wdfn-server/waterdata/camera.py
```diff
@@ -0,0 +1,43 @@
...
+    return json.loads(json_raw)
+
+
+def get_site_camera_data(site_no):
+    json_raw = get_json_camera_data()
+    camera = json_raw_to_dictionary(json_raw)
+    return find_site_in_cameras(site_no, camera)
+
+
+def find_site_in_cameras(site_no, cameras):
```
ayan-usgs:
```suggestion
def find_site_in_cameras(site_no, cameras):
    site_camera = next((sc for sc in cameras if sc['SiteId'] == site_no), {})
    return site_camera
```

wdfn-server/waterdata/camera.py
```diff
@@ -0,0 +1,43 @@
...
+    for site_camera in cameras:
+        if site_no in site_camera['SiteId']:
+            return site_camera
+    return {}
+
+
+def get_monitoring_camera_data(site_no):
+    camera_data = get_site_camera_data(site_no)
+    if 'lastFrame' not in camera_data:
+        return {'exists':False}
```
ayan-usgs: Codacy has probably already complained about wanting a space after the colon.

wdfn-server/waterdata/views.py
```diff
@@ -115,6 +116,11 @@ def monitoring_location(site_no):
                 'questions_link': questions_link,
                 'cooperators': cooperators
             }
+
+            monitoring_camera = get_monitoring_camera_data(site_no)
+            if monitoring_camera:
```
ayan-usgs: I'm not sure this `if` statement is necessary as currently written. `get_monitoring_camera_data` will always return a dictionary with some content, so it'll always evaluate to `True`.

wdfn-server/waterdata/tests/test_camera.py
```diff
@@ -0,0 +1,94 @@
+from unittest import TestCase
+import sys
+sys.path.append("..")
+import camera
+# from .. import camera
```
ayan-usgs: I would prefer that this and other snippets of commented out code (e.g the `print`s) be removed before merging.

wdfn-server/waterdata/camera.py
```diff
@@ -0,0 +1,43 @@
+import json
+import os
+
+# proof of concept. should call service to get
+
+JSON_CAMERA_FILE = os.getcwd()+"/data/monitoring_camera_data.json"
+
+
+def get_json_camera_data():
```
ayan-usgs: Consider adding docstrings to this and the other functions in this module.

wdfn-server/waterdata/tests/test_camera.py
```diff
@@ -0,0 +1,94 @@
+from unittest import TestCase
+import sys
+sys.path.append("..")
+import camera
```
ayan-usgs: I think travis isn't happy about this import. Based on the directory structure, what is currently commented out on line 5 may be more appropriate.

ayan-usgs:
Is this something we actually want to merge or is just a strawman to talk around?

wdfn-server/waterdata/views.py
duselmann: Right, this is because originally I did not return a dictionary and later I found that is was better to return one with the exists state. It could be removed.

wdfn-server/waterdata/tests/test_camera.py
duselmann: This was a last resort. I could not get it to import without the legacy path append style. Py would just not find the module one level up in my workspace.
I had `import ..camera` first and then tried `from .. import camera` and others.

wdfn-server/waterdata/camera.py
duselmann: I tried replacing this and the unit tests fail.

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -68,7 +68,34 @@ <h1>{{ stations[0].station_nm }}</h1>
...
                 {{ components.TimeSeriesComponent(stations[0].site_no) }}
+                {% if monitoring_camera.exists %}
+                <div class="usa-accordion">
+                    <h2 class="usa-accordion__heading">
+                        <button class="usa-accordion__button" aria-expanded="true" aria-controls="ts-a1a">Monitoring Camera</button>
+                    </h2>
+                    <div id="ts-a1a" class="usa-accordion__content">
+                        <div style="float:left">
+                            <a href="{{ monitoring_camera.video  }}" title="click image to open full-size video" alt="lastest camera image thumbnail" target="_blank">
+                                <img width="200" src="{{ monitoring_camera.thumb }}">
```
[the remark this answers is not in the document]
duselmann: I had an alt attribute, I just added it to the wrong tag. It is now it its proper place.

wdfn-server/waterdata/camera.py
duselmann: done.

wdfn-server/waterdata/tests/test_camera.py
duselmann: They are only in a unit test, not in main code.

wdfn-server/waterdata/tests/test_camera.py
mbucknell: Were you trying this from PyCharm? I think you need to set up testing in the Preferences (I know because I always have issues, too).

wdfn-server/waterdata/tests/test_camera.py
mbucknell: Unless there is a compelling reason to leave in commented out code, it should be removed, IMO

jkreft-usgs:
This is something that we do want to merge, since it is an MVP that we can get in front of users (and tell our funders about)

wdfn-server/waterdata/camera.py
ayan-usgs: Okay.

PR #638, opened 2019-09-13
IOW-47 - upstream-downstream sites on map

dsteinich:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/map/index.js
+++ b/assets/src/scripts/components/map/index.js
@@ -1,7 +1,7 @@
 import { select } from 'd3-selection';
 import { createStructuredSelector } from 'reselect';
-import { map as createMap, marker as createMarker, control, layerGroup, icon } from 'leaflet';
-import { TiledMapLayer, dynamicMapLayer, Util, basemapLayer, featureLayer } from 'esri-leaflet/src/EsriLeaflet';
+import { map as createMap, marker as createMarker, control, layerGroup } from 'leaflet';
+import { TiledMapLayer, dynamicMapLayer, Util, basemapLayer } from 'esri-leaflet/src/EsriLeaflet';
 import { link, provide } from '../../lib/redux';
 import config from '../../config';
 import { FLOOD_EXTENTS_ENDPOINT, FLOOD_BREACH_ENDPOINT, FLOOD_LEVEE_ENDPOINT } from '../../flood-data';
@@ -9,6 +9,7 @@ import { hasFloodData, getFloodExtent, getFloodStageHeight } from '../../selecto
 import { Actions } from '../../store';
 import { floodSlider } from './flood-slider';
 import { createLegendControl, createFIMLegend } from './legend';
+import { addNldi } from './nldiMapping';
 
 
 const getLayerDefs = function(layerNo, siteno, stage) {
@@ -139,6 +140,9 @@ const siteMap = function(node, {siteno, latitude, longitude, zoom}) {
 
     // Add a marker at the site location
     createMarker([latitude, longitude]).addTo(map);
+
+    //add nldi layers
+    addNldi(map, legendControl, siteno);
 
     node
         .call(link(updateFloodLayers, createStructuredSelector({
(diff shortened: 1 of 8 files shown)


assets/src/scripts/components/map/index.js
```diff
@@ -140,6 +141,9 @@ const siteMap = function(node, {siteno, latitude, longitude, zoom}) {
     // Add a marker at the site location
     createMarker([latitude, longitude]).addTo(map);
 
+    //add nldi layers
+    addNldi(map, legendControl, siteno);
```
mbucknell: I'd like to see a test to see that this layer gets created. See https://github.com/usgs/waterdataui/blob/master/assets/src/scripts/components/map/index.spec.js#L39, except it you won't look for a .leaflet-marker-pane. I'm not sure what gets created but you should be able to use developer tools to see.

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -0,0 +1,152 @@
...
+ * @param {String} sitno The starting site for navigation
+ */
+export const addNldi = function(map, legendControl, siteno) {
+    const nldiUrl = config.NLDI_SERVICES_ENDPOINT;
+    const featureSource = 'nwissite';
+    const dataSource = 'nwissite';
+    const upstreamNavigation = 'UM';
+    const downstreamNavigation = 'DM';
+    const featureId = 'USGS-' + siteno;
+    const distanceParam = '?distance=322';
```
mbucknell: I'd like to see the distance to be a configuration parameter.  So that would go in server/config.py. Can probably just set it there, maybe Initial_Upstream_Downstream_Distance.

Then you could set this distanceParam=`$distance=${config.NLDI_SERVICES_ENDPOINT}`. Uses template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -0,0 +1,152 @@
...
+
+    const upstreamLineStyle = {
+        'color': upstreamColor,
+        'weight': 5,
+        'opacity':flowLineOpacity
+    };
+
+    const onEachPointFeatureAddPopUp = function(feature, layer) {
+        const uri = feature.properties.uri;
+        const popupText = 'Data Source: ' + feature.properties.source +
```
mbucknell: You might consider using templates literals here too (maybe for each line?)

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -0,0 +1,152 @@
...
+        const lineLayer = geoJson(data, {
+            onEachFeature: onEachLineFeatureAddPopUp,
+            style: style
+            });
+        lineLayer.addTo(map);
+        const features = data.features;
+        features.forEach(function(feature) {
+            allExtents.features.push(feature);
+        });
+//        map.fitBounds(geoJson(allExtents).getBounds());
```
mbucknell: I'd remove this line.

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+    const legendListContainer = select(legendControl.getContainer()).select('.legend-list-container');
+    const nldiLegendList = legendListContainer.append('ul')
+                .attr('id', 'nldi-legend-list')
+                .classed('usa-list--unstyled', true);
+
+    nldiLegendList.append('li')
+                .classed('nldi-legend', true)
+                .html('<li><i style="background:' + upstreamColor + '; width: 16px; height: 16px; float: left; opacity: ' + flowLineOpacity + '; margin-right: 2px;"></i><span>Upstream Flowline</span> </li>'
```
mbucknell: Consider adding each <li> element using D3.

So it would look something like this
`
nldiLegendList.append('li')
     .append('i')
           .attr('style', 'background: ....')
nldiLegendList.append('li')
     .append('i')
           .attr('class', 'fas fa-circle')
           .attr('style', 'style that you want')
`
and so one.

PR #645, opened 2019-09-18
IOW-60 revise and update config for camera

duselmann:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------
IOW-60 revise and update config for camera

Description
-----------
I was unaware of the full config and application settings nature of the application.
The prior pull request did not work because of inaccurate path to data file.
It also would have failed on the MONITORING_CAMERA_ENABLE flag missing from the conf.

After making a pull request
---------------------------

### DIFF
--- a/wdfn-server/config.py
+++ b/wdfn-server/config.py
@@ -19,6 +19,7 @@ STATE_COUNTY_PAGES_ENABLED = True
 EMBED_IMAGE_FEATURE_ENABLED = True
 COOPERATOR_LOOKUP_ENABLED = True  # may also be set to a list of district codes
 MONITORING_CAMERA_ENABLED = True
+MONITORING_CAMERA_PATH    = 'wdfn-server/data/monitoring_camera_data.json'
 
 DEBUG = False
 
--- a/wdfn-server/waterdata/tests/test_camera.py
+++ b/wdfn-server/waterdata/tests/test_camera.py
@@ -4,7 +4,9 @@ Tests for camera.py module
 from unittest import TestCase
 from .. import camera
 
-camera.JSON_CAMERA_FILE = "wdfn-server/data/monitoring_camera_data_test.json"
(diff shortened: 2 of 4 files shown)


wdfn-server/waterdata/tests/test_camera.py
```diff
@@ -4,7 +4,9 @@
 from unittest import TestCase
 from .. import camera
 
-camera.JSON_CAMERA_FILE = "wdfn-server/data/monitoring_camera_data_test.json"
+JSON_CAMERA_TEST_FILE = "wdfn-server/data/monitoring_camera_data_test.json"
+with open(JSON_CAMERA_TEST_FILE, 'r') as json_file:
+     camera.MONITORING_CAMERA_RAW = json_file.read()
```
ayan-usgs: Not a huge deal, but I'm wondering if it would be cleaner to mock out this constant rather than replacing it.

wdfn-server/config.py
```diff
@@ -19,6 +19,7 @@
 EMBED_IMAGE_FEATURE_ENABLED = True
 COOPERATOR_LOOKUP_ENABLED = True  # may also be set to a list of district codes
 MONITORING_CAMERA_ENABLED = True
+MONITORING_CAMERA_PATH    = 'wdfn-server/data/monitoring_camera_data.json'
```
ayan-usgs: Nit-picky, but I'd remove the extra spaces.

wdfn-server/waterdata/tests/test_camera.py
duselmann: I am all for it. Anything in particular you had in mind?

wdfn-server/config.py
duselmann: You prefer the '=' not to line up?

duselmann:
While there are a couple issues we could revise, there is a 500 we would like to try to clear up. I will post these suggestions in the story. Further, this is suppose to proof of concept. Any or all of this could vastly change in the actual release.

PR #689, opened 2019-10-04
Iow 33 - Updated to this project to use the latest wdfn-viz

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Updated project using the latest guidance in wdfn-viz and USWDS.

After making a pull request
---------------------------


mbucknell:
Regarding the codacy review. I added stylelintrc to this project and set the rules to be the same as USWDS. I've already adjusted a few. I'm inclined at this point to leave as is with the few rule violations. We may want to figure whether we want to keep the rules being violated or not but I think this is good for now.

assets/package.json
```diff
@@ -75,11 +75,11 @@
     "rollup-plugin-node-resolve": "^5.2.0",
     "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-uglify": "^6.0.3",
-    "uglify-js": "^3.6.0",
-    "uglifycss": "0.0.29",
```
mbucknell: using cssnano rather tha uglifycss. This actually compresses better and is recommended by USWDS.

assets/package.json
```diff
@@ -75,11 +75,11 @@
...
-    "wdfn-viz": "1.0.0"
+    "stylelint": "^10.1.0",
+    "stylelint-scss": "^3.11.1"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-free": "5.11.2",
```
mbucknell: Moved this dependency and wdfn-viz from dev-dependencies. Not sure why we put them there. The are really application dependencies and all of the dependencies have to be loaded in order to build.

assets/package.json
```diff
@@ -41,8 +41,9 @@
   },
   "homepage": "https://github.com/usgs/waterdataui#readme",
   "devDependencies": {
-    "@fortawesome/fontawesome-free": "5.11.2",
     "autoprefixer": "^9.6.1",
+    "css-mqpacker": "^7.0.0",
+    "cssnano": "^4.1.10",
```
ayan-usgs: Is this automatically being pulled in by postcss? The cssnano docs mention that it needs to be configured in a `postcss.config.js` file. :thinking:

mbucknell: It is in postcss.config.js which was added to this repo.

PR #705, opened 2019-10-09
Iow 66 make median line optional

ssoper-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------
JIRA: https://internal.cida.usgs.gov/jira/browse/IOW-66

Description
-----------
Added a feature toggle to enable/disable median line on hydrograph.

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/graph-controls.js
```diff
@@ -0,0 +1,82 @@
...
+import { dispatch, link } from '../../lib/redux';
+import { audibleUI } from './audible';
+import { getCurrentVariableMedianStatistics } from '../../selectors/median-statistics-selector';
+import { isVisibleSelector, currentVariableTimeSeriesSelector } from './time-series';
+
+/*
+ * Create the show audible toggle, last year toggle, and median toggle for the time series graph.
+ * @param {Object} elem - D3 selection
+ */
+export const graphControls = function(elem) {
```
mbucknell: Could we call this drawGraphControls? I think that is the pattern we have established and I like using verbs in function names.

assets/src/scripts/components/hydrograph/graph-controls.js
```diff
@@ -0,0 +1,82 @@
...
+
+    const graphControlDiv = elem.append('ul')
+        .classed('usa-fieldset', true)
+        .classed('usa-list--unstyled', true)
+        .classed('graph-controls-container', true);
+
+    graphControlDiv.append('li')
+        .call(audibleUI);
```
mbucknell: I had forgotten about the audibleUI. This is another reason I would like this module to separate out the creating of the compare and median toggles. But in the interest of getting something that works, this is fine.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        }
+    };
+
+    let graphNode;
+
+    beforeEach(() => {
+        let body = select('body');
+        let component = body.append('div')
+            .attr('id', 'hydrograph');
+        component.append('div').attr('class', 'loading-indicator-container');
```
mbucknell: I think the only DOM element you need is the component to create the graphControl in.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
+import {Actions, configureStore} from '../../store';
+import {attachToNode} from './index';
```
mbucknell: See below. You don't need this import. I'm a little surprised that the test works with this imported.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        it('should be enabled if there are last year data', () => {
+            expect(select('#last-year-checkbox').property('disabled')).toBeFalsy();
+        });
+
+        it('should be disabled if there are no last year data', () => {
+            store.dispatch(Actions.setCurrentVariable('45807190'));
+            expect(select('#last-year-checkbox').property('disabled')).toBeTruthy();
+        });
+
+        it('Should render one lines', () => {
```
mbucknell: These should go back into the index.spec.js test. This does not test that the checkbox is rendered correctly when the state changes. This is testing code in index.js and does not belong here.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
...
+            expect(selectAll('#ts-compare-group .line-segment').size()).toBe(1);
+        });
+
+        it('Should remove the lines when removing the compare time series', () => {
```
mbucknell: Same as line 202 comment

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
...
+        it('should be enabled if there are median statistics data', () => {
+            expect(select('#median-checkbox').property('disabled')).toBeFalsy();
+        });
+
+        it('should be disabled if there are no median statistics data', () => {
+            store.dispatch(Actions.setCurrentVariable('45807190'));
+            expect(select('#median-checkbox').property('disabled')).toBeTruthy();
+        });
+
+        it('Should render one lines', () => {
```
mbucknell: Same as comment on line 202 except this can be added to index.js.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
```diff
@@ -0,0 +1,243 @@
...
+            expect(selectAll('#median-points .median-data-series').size()).toBe(1);
+        });
+
+        it('Should remove the lines when removing the median statistics data', () => {
```
mbucknell: Same as comment on line 202 except this can be added to index.js

assets/src/scripts/components/hydrograph/graph-controls.js
ssoper-usgs: Yeah that's a great idea!

assets/src/scripts/components/hydrograph/graph-controls.js
ssoper-usgs: I agree, I will look at splitting this logic out if there's time at the end of the sprint.  If I don't make it, I'll make a jira task.

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

assets/src/scripts/components/hydrograph/graph-controls.spec.js
ssoper-usgs: done!

PR #762, opened 2019-12-05
Iow 57

ssoper-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------
JIRA task: https://internal.cida.usgs.gov/jira/browse/IOW-57

Description
-----------
Legend on map should be collapsible if the upstream/downstream and/or the Flood layers are available.

After making a pull request
---------------------------


assets/src/scripts/components/map/index.js
```diff
@@ -8,8 +8,15 @@ import { FLOOD_EXTENTS_ENDPOINT, FLOOD_BREACH_ENDPOINT, FLOOD_LEVEE_ENDPOINT } f
 import { hasFloodData, getFloodExtent, getFloodStageHeight } from '../../selectors/flood-data-selector';
 import { Actions } from '../../store';
 import { floodSlider } from './flood-slider';
-import { createLegendControl, createFIMLegend } from './legend';
-import { addNldi } from './nldiMapping';
+import { createLegendControl, createFIMLegend, createNldiLegend } from './legend';
+import { addNldiLayers} from './nldiMapping';
+import {
+    hasNldiData,
```
mbucknell: Nit. Our pattern has been to put all on the same line. If more than one line is required, then continue on to the next line (see https://github.com/usgs/waterdataui/blob/master/assets/src/scripts/components/hydrograph/index.js for examples)

assets/src/scripts/components/map/index.spec.js
```diff
@@ -138,4 +139,192 @@ describe('map module', () => {
...
+
+        it('Should create a leaflet map within the mapNode with', () => {
+            expect(select(mapNode).selectAll('.leaflet-container').size()).toBe(1);
+        });
+
+        it('Should create a marker layer', () => {
+            expect(select(mapNode).selectAll('.leaflet-marker-pane img').size()).toBe(1);
+        });
+
+        it('Should not create NLDI layers', () => {
```
mbucknell: Another nit. Maybe string should be 'Should not create an overlay layers'. The FIM layers are also in the overlay pane.

assets/src/scripts/components/map/index.spec.js
```diff
@@ -138,4 +139,192 @@ describe('map module', () => {
...
+        it('Should create NLDI layers', () => {
+            expect(select(mapNode).selectAll('.leaflet-overlay-pane svg g').size()).toBe(1);
+            expect(select(mapNode).selectAll('.leaflet-overlay-pane svg g path').size()).toBe(8);
+        });
+
+        it('Should create a NLDI Legend', () => {
+            expect(select(mapNode).select('#nldi-legend-list').size()).toBe(1);
+        });
+
+    });
```
mbucknell: Nice addition to tests!

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -1,7 +1,12 @@
 import { geoJson, circleMarker } from 'leaflet';
-import { get } from '../../ajax';
-import config from '../../config';
-import { select } from 'd3-selection';
+import {
```
mbucknell: Same as above. Let's try to get all on a single line.

assets/src/scripts/components/map/nldiMapping.js
```diff
@@ -1,7 +1,12 @@
...
+    markerFillColor,
```
mbucknell: I think these belong in here and then legend can import them. What do you think? I think that makes it clearer that these are nldi style settings.

ssoper-usgs: Yeah I think that's a better approach, I'll switch this around.

assets/src/scripts/components/map/nldiMapping.js
ssoper-usgs: Will do.

assets/src/scripts/components/map/index.spec.js
ssoper-usgs: Yup - good point, I'll reword this.

PR #792, opened 2019-12-18
Iow 176 - Image server: support periods

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Image server now supports the parameter period=xxD where xx represents the number of days of data to return. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/index.js
+++ b/assets/src/scripts/components/hydrograph/index.js
@@ -8,12 +8,14 @@ import { select } from 'd3-selection';
 import { DateTime } from 'luxon';
 import { createStructuredSelector } from 'reselect';
 
+import { dispatch, link, provide } from '../../lib/redux';
+
 import { addSVGAccessibility } from '../../accessibility';
 import config from '../../config';
-import { dispatch, link, provide } from '../../lib/redux';
-import { getTimeSeriesCollectionIds, isLoadingTS, hasAnyTimeSeries } from '../../selectors/time-series-selector';
+import { isLoadingTS, hasAnyTimeSeries } from '../../selectors/time-series-selector';
 import { Actions } from '../../store';
 import { callIf, mediaQuery } from '../../utils';
+
 import { appendAxes, axesSelector } from './axes';
 import { cursorSlider } from './cursor';
 import { lineSegmentsByParmCdSelector, currentVariableLineSegmentsSelector, MASK_DESC, HASH_ID,
@@ -24,8 +26,7 @@ import { drawSimpleLegend, legendMarkerRowsSelector } from './legend';
 import { drawMethodPicker } from './method-picker';
 import { plotSeriesSelectTable, availableTimeSeriesSelector } from './parameters';
 import { xScaleSelector, yScaleSelector, timeSeriesScalesByParmCdSelector } from './scales';
-import { allTimeSeriesSelector, isVisibleSelector, titleSelector, descriptionSelector,
-    hasTimeSeriesWithPoints } from './time-series';
+import { allTimeSeriesSelector, isVisibleSelector, titleSelector, descriptionSelector } from './time-series';
 import { createTooltipFocus, createTooltipText } from './tooltip';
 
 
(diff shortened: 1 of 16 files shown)


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -481,33 +487,31 @@ const dateRangeControls = function(elem, siteno) {
...
-                    .text('No current time series data available for this site');
+            .attr('class', 'usa-alert-body')
+            .append('p')
+            .attr('class', 'usa-alert-text')
+            .text(message);
     }
 };
 
-export const attachToNode = function (store, node, {siteno, parameter, compare, cursorOffset, interactive = true} = {}) {
+export const attachToNode = function (store, node, {siteno, parameter, compare, period, cursorOffset, showOnlyGraph = false} = {}) {
```
mbucknell: In addition to changes required for the period, I reordered the code to put fetching before the rendering code. Also reordered the rendering code so it's clearer what is being rendered. 

I think this module could use some decomposition (for instance moving the date controls into their own module and perhaps moving the chart into its own module) so that the index module is basically using functions from submodules. I think it would make testing easier. I am going to create a tech-debt JIRA to address this.

assets/src/scripts/components/hydrograph/index.spec.js
```diff
@@ -163,9 +163,12 @@ describe('Hydrograph charting module', () => {
         component.append('div').attr('class', 'provisional-data-alert');
 
         graphNode = document.getElementById('hydrograph');
+
+        jasmine.Ajax.install();
```
mbucknell: We were actually making NWIS service calls as part of the test suite. This may be one reason our travis builds sometimes failed. By adding this, all ajax calls are stubbed.

assets/src/scripts/store/index.js
```diff
@@ -435,7 +459,7 @@ export const configureStore = function (initialState) {
                 median: false
             },
             currentDateRange: 'P7D',
-            requestedTimeRange: null,
+            customTimeRange: null,
```
mbucknell: I was getting confused by requestedTimeRange and requestDT in the series part of the state. Changed this to customTimeRange because everything is requested and it seems more descriptive.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -481,33 +487,31 @@ const dateRangeControls = function(elem, siteno) {
...
+    const nodeElem = select(node);
```
ssoper-usgs: This makes the code easier to read, thanks!

assets/src/scripts/components/hydrograph/index.spec.js
ssoper-usgs: Nice find

assets/src/scripts/store/index.spec.js
```diff
@@ -403,6 +402,72 @@ describe('Redux store', () => {
...
+                p.then(() => {
+                    expect(Actions.setCurrentVariable).toHaveBeenCalledWith(45807197);
+                    expect(Actions.addSeriesCollection).toHaveBeenCalled();
+                    expect(Actions.addSeriesCollection.calls.argsFor(0)[0]).toEqual('current:custom:00060');
+                    expect(Actions.removeTimeSeriesLoading).toHaveBeenCalledWith(['current:custom:00060']);
+                    done();
+                });
+            });
+
+            it('Should clear the data for and remote the time series loader for bad data', (done) => {
```
ssoper-usgs: `remove` rather than `remote`

graph-server/src/renderer/index.js
```diff
@@ -5,13 +5,14 @@ const PAST_SERVICE_ROOT = process.env.PAST_SERVICE_ROOT || 'https://nwis.waterse
 const STATIC_ROOT = process.env.STATIC_ROOT || 'https://waterdata.usgs.gov/nwisweb/wsgi/static';
 
 
-const renderToRespone = function (res, {siteID, parameterCode, compare}) {
+const renderToRespone = function (res, {siteID, parameterCode, compare, period}) {
     const componentOptions = {
         siteno: siteID,
         parameter: parameterCode,
         compare: compare,
+        period: period,
```
ssoper-usgs: Cool.

PR #807, opened 2019-12-27
Iow 217 break up control and charting code into smaller modules

ssoper-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------
JIRA: https://internal.cida.usgs.gov/jira/browse/IOW-217

Description
-----------
Break out date control and time series graph logic and tests into their own modules.

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/time-series-graph.spec.js
```diff
@@ -0,0 +1,388 @@
...
+
+        jasmine.Ajax.install();
+    });
+
+    afterEach(() => {
+        jasmine.Ajax.uninstall();
+        select('#hydrograph').remove();
+    });
+
+    it('empty graph displays warning', () => {
```
ssoper-usgs: Lines 175 to... 188 or possibly a few of the early tests might be considered initialization tests and could maybe stay in the index.spec.

assets/src/scripts/components/hydrograph/index.spec.js
```diff
@@ -150,7 +149,7 @@ const TEST_STATE = {
 };
 
 
-describe('Hydrograph charting module', () => {
+describe('Loading indicators and data alerts', () => {
```
ssoper-usgs: The remaining tests might belong in the time-series-graph.spec, but the loading indicators seem to be used in other areas of the page so I left them here for the time being.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -333,172 +62,14 @@ const controlDisplay = function(elem, showElem) {
     elem.attr('hidden', showElem ? null : true);
 };
 
-const loadingIndicator = function(elem, {showLoadingIndicator, sizeClass}) {
+export const loadingIndicator = function(elem, {showLoadingIndicator, sizeClass}) {
```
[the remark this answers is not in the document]
mbucknell: I think that is an excellent idea. In fact, it really is more of a utility function. I'm tempted to put it under components even though it is used within other "components".

assets/src/scripts/components/hydrograph/date-controls.spec.js
```diff
@@ -0,0 +1,246 @@
+import { select } from 'd3-selection';
+import {Actions, configureStore} from '../../store';
+import {attachToNode} from './index';
```
mbucknell: You should test the exported function from the date-controls module (drawDateRangeControls) rather than calling attachToNode. You should be able to modify the tests below and potentially add additional tests because may only need to mock the part of the state that you need need to test the rendering code with.

assets/src/scripts/components/hydrograph/index.spec.js
mbucknell: Yes, there are actually two loading indicators on the page so it's appropriate to leave this here

assets/src/scripts/components/hydrograph/time-series-graph.js
```diff
@@ -0,0 +1,280 @@
+import { extent } from 'd3-array';
+import { line as d3Line, curveStepAfter } from 'd3-shape';
+import {createTooltipFocus, createTooltipText} from './tooltip';
```
mbucknell: Let's clean up the import ordering. Put a space between library level imports (the d3-*) and then group together imports that are from the same directory, and put the imports in that group in alphabetical order. Makes it easier to find things.

assets/src/scripts/components/hydrograph/time-series-graph.spec.js
```diff
@@ -0,0 +1,388 @@
+import { select, selectAll } from 'd3-selection';
+import {attachToNode, timeSeriesLegend} from './index';
```
mbucknell: Same comment as for the date-controls.spec.js.

assets/src/scripts/components/hydrograph/time-series-graph.spec.js
mbucknell: I agree that they belong there. I think it will be clearer what belongs here once you call the exporting function in time-series-graph.js

assets/src/scripts/components/hydrograph/time-series-graph.spec.js
```diff
@@ -0,0 +1,388 @@
...
+        it('should not have tooltips for the select series table when the screen is large', () => {
+            store.dispatch(Actions.resizeUI(800, 800));
+            expect(selectAll('table .tooltip-table').size()).toBe(0);
+        });
+    });
+
+    //TODO: Consider adding a test which checks that the y axis is rescaled by
+    // examining the contents of the text labels.
+
+    describe('legends should render', () => {
```
mbucknell: This doesn't belong here at all because the legend code is not part of the time-series-graph rendering.

assets/src/scripts/components/hydrograph/date-controls.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+    afterEach(() => {
+        div.remove();
+    });
+
+    it('Expects the date range controls to be created', () => {
+        let dateRangeContainer = select('#ts-daterange-select-container');
+        let customDateDiv = select('div#ts-customdaterange-select-container');
+
+        expect(dateRangeContainer.size()).toBe(1);
```
mbucknell: It's a little weird that you would use .size(). The select function only ever returns one element so it either finds it or it doesn't. I believe if it doesn't find the element, the variable would be null. Maybe that's a better test to use.

assets/src/scripts/components/hydrograph/date-controls.spec.js
```diff
@@ -0,0 +1,235 @@
...
+    it('Expects an alert to be thrown if custom dates are not provided.', () => {
+         let submitButton = select('#custom-date-submit');
+         submitButton.dispatch('click');
+
+         let customDateAlertDiv = select('#custom-date-alert');
+         expect(customDateAlertDiv.attr('hidden')).toBeNull();
+         expect(customDateAlertDiv.select('p').text()).toEqual('Both start and end dates must be specified.');
+    });
+
+    it('Expects and alert to be thrown if the end date is earier than the start date.', () => {
```
mbucknell: Nit typo "and" should be "an".

assets/src/scripts/components/loadingIndicator.js
```diff
@@ -0,0 +1,7 @@
+export const loadingIndicator = function(elem, {showLoadingIndicator, sizeClass}) {
```
mbucknell: Let's change the name of this module to loading-indictor.js to follow the js module naming convention already established. Same with the spec file.

mbucknell: Can we call this drawLoadingIndicator? That's the pattern in the other modules like date-range-control.js

assets/src/scripts/components/hydrograph/date-controls.spec.js
ssoper-usgs: Mary and I did a little sleuthing and determined this is an ok way to check for the existence of the date control container given that we are using d3.select.

PR #836, opened 2020-01-22
Update d3 redux

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Updated the lib/redux module to simplify as was done in the NGWMN site pages and renamed to d3-redux to make it clear that this is not part of the standard redux library.

After making a pull request
---------------------------


assets/package.json
```diff
@@ -93,6 +93,7 @@
     "luxon": "1.21.3",
     "matchmedia-polyfill": "0.3.2",
     "normalizr": "3.4.1",
+    "raf-throttle": "2.0.3",
```
mbucknell: Added the throttling to prevent unnecessary rerenderings

assets/src/scripts/components/hydrograph/audible.spec.js
```diff
@@ -81,11 +78,13 @@ describe('audibleUI', () => {
         expect(store.getState().timeSeriesState.audiblePlayId).not.toBeNull();
     });
 
-    it('Expects the store to have a null playId if the  button is clicked after clicking the once', () => {
+    it('Expects the store to have a null playId if the  button is clicked after clicking the once', (done) => {
         container.select('button').dispatch('click');
-        container.select('button').dispatch('click');
-
-        expect(store.getState().timeSeriesState.audiblePlayId).toBeNull();
+        window.requestAnimationFrame(() => {
```
mbucknell: Anytime you check the state of the DOM after an update to Redux, you now need to wait until the next animation frame due to the throttling. You will see this pattern repeat in the tests.

assets/package.json
JanellFry-usgs: Guess I need to look up throttling, are you trying to speed things up?

mbucknell: I'm trying to prevent the javascript from trying to rerender except when it is time for the browser to do so. It should make the browser less busy and maybe speed up rendering of the page. You might want to look at window.requestAnimationFrame as well.

PR #868, opened 2020-02-06
IOW-241 First cut at adding a Networks page

JanellFry-usgs:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


wdfn-server/waterdata/services/ogc.py
```diff
@@ -0,0 +1,35 @@
...
+    response = execute_get_request(url)
+
+    if response.status_code != 200:
+        return []
+    try:
+        resp_json = response.json()
+    except ValueError:
+        return []
+    else:
+        if network_cd == 'ALL':
```
mbucknell: I think the whole object should be returned rather than just the collection part. This service method may be used for different purposes in the future and returning the whole object make the method more reusable. Then we can eliminate this else clause.

wdfn-server/waterdata/views.py
```diff
@@ -191,6 +191,28 @@ def hydrological_unit_locations(huc_cd):
     return hydrological_unit(huc_cd, show_locations=True)
 
 
+@app.route('/networks/', defaults={'network_cd': 'ALL'}, methods=['GET'])
+@app.route('/networks/<network_cd>/', methods=['GET'])
```
mbucknell: This is really part of the next JIRA for networks but okay.

wdfn-server/waterdata/views.py
```diff
@@ -191,6 +191,28 @@ def hydrological_unit_locations(huc_cd):
...
+def networks(network_cd):
+    """
+    Network unit view
+    :param network_cd: ID for this network
+    """
+
+    # Grab the Network info
+    collection = ogc.get_networks(network_cd)
+
+    http_code = 200 if (network_cd != '') else 404
```
mbucknell: I think we should return 404 if collection is an empty dictionary.

wdfn-server/waterdata/views.py
```diff
@@ -191,6 +191,28 @@ def hydrological_unit_locations(huc_cd):
...
+
+    return render_template(
+        'networks.html',
+        http_code=http_code,
```
mbucknell: I don't think we should be passing the http_code as part of the context data. The Flask framework handles non 200 http codes automatically.

wdfn-server/waterdata/services/ogc.py
```diff
@@ -0,0 +1,35 @@
...
+    if not app.config['NETWORK_PAGES_ENABLED']:
+        return []
+
+    url = app.config['OGC_SERVICE'].format('json')
+
+    if network_cd != 'ALL':
+        url = url +'/'+network_cd+'/'
+    url = url + '?f=json'
+
+    #print('services: ogc.py: get_networks(): ' + url)
```
mbucknell: Please remove commented out code.

wdfn-server/waterdata/views.py
```diff
@@ -191,6 +191,28 @@ def hydrological_unit_locations(huc_cd):
     return hydrological_unit(huc_cd, show_locations=True)
 
 
+@app.route('/networks/', defaults={'network_cd': 'ALL'}, methods=['GET'])
```
mbucknell: Should default to None following the pattern for states.

wdfn-server/waterdata/templates/networks.html
```diff
@@ -0,0 +1,47 @@
...
+        {% if http_code == 200 %}
+            <h1>Networks</h1>
+
+            {% if network_cd == 'ALL' %}
+                <p>Select a Network</p>
+                <table class="usa-table">
+                    <thead>
+                        <tr>
+                            <th scope="col">Network</th>
+                            <th scope="col">Type</th>
```
mbucknell: I don't think itemType is useful as it will always be 'feature'.

wdfn-server/waterdata/templates/networks.html
```diff
@@ -0,0 +1,47 @@
...
+                                <td>{{ network.itemType }}</td>
+                                <td>{{ network.title }}</td>
+                                <td>{{ network.description }}</td>
+                            </tr>
+                        {% endfor %}
+                    </tbody>
+                </table>
+
+            {% else %}
+                {% if collection.description %}
```
mbucknell: So this kind of started the next ticket, but I it's okay.

wdfn-server/waterdata/tests/test_views.py
```diff
@@ -142,6 +142,48 @@ def test_locations_list(self, client):
...
+        assert response.status_code == 200
+
+    @mock.patch('waterdata.services.ogc.get_networks')
+    def test_invalid_network(self, network_mock):
+        network_mock.return_value.status_code = 200
+        network = 'monitoring-locations-invalid'
+        url = '/networks/{}/'.format(network)
+        response = self.app_client.get(url)
+        # Assert an invalid network in gui does not 404
+        assert response.status_code == 200
```
mbucknell: So we are getting ahead of this JIRA as this part of the next but I think it should be a 404.

wdfn-server/waterdata/services/ogc.py
```diff
@@ -0,0 +1,35 @@
...
+from waterdata.utils import execute_get_request
+
+
+def get_networks(network_cd):
+    """
+    Gets the network data from a json file
+    :param network_cd: collections-id
+    """
+    # Handle feature flag for network pages
+    if not app.config['NETWORK_PAGES_ENABLED']:
```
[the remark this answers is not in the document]
JanellFry-usgs: That's fine if you don't want a feature flag here, I was just following the get_cooperators() pattern in services/sifta.py here, seems to have one there.  Should we not have a flag?

wdfn-server/waterdata/services/ogc.py
```diff
@@ -0,0 +1,35 @@
...
+        return []
+
+    url = app.config['OGC_SERVICE'].format('json')
```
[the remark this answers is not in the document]
JanellFry-usgs: Good call, probably don't need that...

wdfn-server/waterdata/services/ogc.py
JanellFry-usgs: I was noticing that the return is different if you are requesting a specific network as opposed to all networks.  The collection part in the json is not there if you only ask for a specific network.  I wonder if the call itself should be changed to look the same regardless of whether you're asking for one network or all?

wdfn-server/waterdata/services/ogc.py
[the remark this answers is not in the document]
mbucknell: That one is a little different in that it gets cooperator information either from a service (if enabled) or a static file. In our case, we feature flag the view which you are doing. Now that I double check the JIRA I see that a feature flag is not part of the acceptance criteria. You should verify with @jkreft-usgs  whether one is needed

wdfn-server/waterdata/templates/networks.html
JanellFry-usgs: Sure, I can remove the itemType column if you want.

wdfn-server/waterdata/services/ogc.py
mbucknell: That's what I'm thinking. Just return the entire response which will always be a dictionary.

wdfn-server/waterdata/services/ogc.py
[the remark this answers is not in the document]
jkreft-usgs: We can probably go without a feature flag on this one.  Greg will have to do an apache mapping before it can be publicly mapped on waterdata.usgs.gov anyways.

PR #892, opened 2020-02-21
IOW-304 First-cut of brush handles and acknowledgement tip on mouseover of brush handles

JanellFry-usgs:
I've been trying different things, but this is the only one that's worked so far.

Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/components/hydrograph/graph-brush.js
+++ b/assets/src/scripts/components/hydrograph/graph-brush.js
@@ -21,6 +21,9 @@ export const drawGraphBrush = function(container, store) {
         }
         const xScale = getBrushXScale('current')(store.getState());
         const brushRange = event.selection || xScale.range();
+
+        console.log('brushRange:'+brushRange)
+
         // Only about the main hydrograph when user is done adjusting the time range.
         if (event.sourceEvent.type === 'mouseup' || event.sourceEvent.type === 'touchend') {
             store.dispatch(Actions.setHydrographXRange(brushRange.map(xScale.invert, xScale)));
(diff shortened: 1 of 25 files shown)


assets/src/scripts/components/hydrograph/graph-brush.js
```diff
@@ -70,11 +72,25 @@ export const drawGraphBrush = function(container, store) {
...
             const group = svg.append('g').attr('class', 'brush')
                 .attr('transform', `translate(${layout.margin.left},${layout.margin.top})`);
+
             graphBrush.extent([[0, 0], [layout.width - layout.margin.right, layout.height - layout.margin.bottom]]);
+
+            // Creates the brush
             group.call(graphBrush);
+
+            // Fill & round corners of brush handles, set mouseover highlight
+            svg.selectAll('.handle').style('fill', '#022F54').style('opacity', .5)
```
mbucknell: I would like to see the style defined in the Sass style sheets. You could put the style for this in https://github.com/usgs/waterdataui/blob/master/assets/src/styles/components/hydrograph/_app.scss. I would also like to stick with USWDS color tokens: https://designsystem.digital.gov/design-tokens/color/system-tokens/. For the transitions below you could add a class during the first transition which will have the style you want for the transition and then remove that class.

JanellFry-usgs: I created a brush-handle-fill style class.  I tried to also create a separate highlight class and toggle between the two within the transition, but it didn't seem to like that, but I choose two colors from the "Blue warm" category.

assets/src/scripts/components/hydrograph/graph-brush.js
```diff
@@ -70,11 +72,24 @@ export const drawGraphBrush = function(container, store) {
...
+
             graphBrush.extent([[0, 0], [layout.width - layout.margin.right, layout.height - layout.margin.bottom]]);
+
+            // Creates the brush
             group.call(graphBrush);
+
+            // Fill & round corners of brush handles, set mouseover highlight
+            svg.selectAll('.handle').classed('brush-handle-fill', true)
+                .attr('rx',15).attr('ry',15)  // Round corners
+                .on('mouseover', function(){
```
mbucknell: Since changing the class won't work, let's just go ahead and put the fill style in the code (i.e. remove fill: #345d96 from the style sheet). I also take back what I said about using USWDS color tokens. It doesn't look like these can be easily used to set the fill property. But the colors you picked are fine.

PR #895, opened 2020-02-25
Iow 210 - State of hydrograph is saved for bookmarking/linking

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
State of hydrograph is saved in the hash portion of the browser url and can be restored

Description
-----------
Please note that this PR changes how data is fetched so that for the graph server only data that is needed is fetched. Also by using promises, no hydrograph rendering is attempted until the data has been fetched.

After making a pull request
---------------------------


assets/src/scripts/components/hydrograph/index.js
```diff
@@ -34,16 +35,22 @@ const controlDisplay = function(elem, showElem) {
...
+/*
+ * Renders the hydrograph on the node element using the Redux store for state information. The siteno, latitude, and
+ * longitude are required parameters. All others are optional and are used to set the initial state of the hydrograph.
+ */
 export const attachToNode = function (store,
                                       node,
                                       {
                                           siteno,
-                                          parameter,
+                                          latitude,
```
mbucknell: Now we are passing in the location to the hydrograph rendering function. This requires that the location is fetched in the graph-server prior to rendering. This makes it explicit where the time zone is fetched rather than extracting the location from the IV service result and then fetching it.

assets/src/scripts/components/hydrograph/time-series.js
```diff
@@ -38,71 +36,6 @@ export const TEMPERATURE_PARAMETERS = {
...
 
-/**
- * Returns a selector that, for a given tsKey and period:
- * Selects all time series data. If period is null then, the currentDateRange is used
- * @param  {String} tsKey   Time-series key
- * @param {String} period or null date range using ISO-8601 duration;
- * @param  {Object} state   Redux state
- * @return {Object} - Keys are tsID, values are time-series data
- */
-export const getTimeSeriesForTsKey = memoize((tsKey, period) => createSelector(
```
mbucknell: The selectors that were removed are needed to manage the url-params. That module is in the base scripts directory and I did not want it to depend on module that should only be used by hydrograph. We could probably stand some selector clean-up but that will need to wait.

assets/src/scripts/store/index.js
```diff
@@ -81,22 +80,15 @@ export const Actions = {
             const currentState = getState();
             const requestKey = getTsRequestKey('current', 'P7D')(currentState);
             dispatch(Actions.addTimeSeriesLoading([requestKey]));
-
-            return getTimeSeries({sites: [siteno], params}).then(
+            return  getTimeSeries({sites: [siteno], params}).then(
```
ssoper-usgs: might be an extra space between return and get here

assets/src/scripts/url-params.js
```diff
@@ -0,0 +1,53 @@
...
+
+/*
+ * Return {String} hash part of url minus the leading '#'.
+ */
+export const getParamString = function() {
+    const hash = window.location.hash;
+    return hash.length ? hash.substring(1) : '';
+};
+
+export const renderTimeSeriesUrlParams = function(store) {
```
ssoper-usgs: This is pretty neat!

assets/src/scripts/url-params.spec.js
```diff
@@ -0,0 +1,178 @@
...
+                    ...TEST_STATE.timeSeriesState,
+                    showSeries: {
+                        compare: true
+                    }
+                }
+            });
+            renderTimeSeriesUrlParams(store);
+
+            expect(window.location.hash).toContain('parameterCode=00010');
+            expect(window.location.hash).toContain('compare=true');
```
ssoper-usgs: These comparisons between tests make it really clear what the url does/does not have.

assets/src/scripts/url-params.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+            expect(window.location.hash).toContain('parameterCode=00010');
+            expect(window.location.hash).not.toContain('compare=true');
+            expect(window.location.hash).not.toContain('period');
+            expect(window.location.hash).not.toContain('startDT');
+            expect(window.location.hash).not.toContain('endDT');
+            expect(window.location.hash).toContain('timeSeriesId=69928');
+        });
+    });
+});
```
ssoper-usgs: great tests

PR #899, opened 2020-02-29
Add upstream basin to nldi map adjusted

gpetrochenkov-usgs:
All tests pass

Added upstream basin to nldi map on monitoring location page.
IOW-365
Fixed code quality issues.

Thank you.


assets/src/scripts/web-services/nldi-data.js
```diff
@@ -29,21 +29,26 @@ const fetchNldiData = function(nldiQuery, siteno) {
 
 // nldi webservice calls
 export const fetchNldiUpstreamSites = function(siteno) {
-    const upstreamSitesQuery = nldiUrl + '/'+ featureSource + '/' + usgs + siteno + '/navigate/' + upstreamNavigation + '/' + dataSource + distanceParam;
+    const upstreamSitesQuery = nldiUrl + '/' + featureSource + '/' + usgs + siteno + '/navigate/' + upstreamNavigation + '/' + dataSource + distanceParam;
```
mbucknell: It would be lovely to change these to using template literals as long as you are in here. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

PR #905, opened 2020-03-04
Fixed compare time series brush disappearance

gpetrochenkov-usgs:
All tests pass

IOW-299
Fixed compare time series disappearance when adjusting the brush on the monitoring location pages.

Thanks!


assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -75,7 +76,14 @@ export const getXScale = memoize((kind, tsKey) => createSelector(
         if (kind === 'BRUSH') {
             timeRange = requestTimeRange;
         } else {
-            timeRange = hydrographXRange ? hydrographXRange : requestTimeRange;
+            if (hydrographXRange && Object.keys(hydrographXRange).length > 0) {
```
mbucknell: This seems like we are fixing a symptom instead of the problem. The issue is that getRequestTimeRange appears to be returning the wrong range. The comment of getRequestTimeRange  is 
```
 @return {Object} with start and end {Number} properties that contain the range of the data requested in universal time or null if the store does not contain a query for the tsKey request```

File: assets/src/scripts/components/hydrograph/graph-brush.js [hunk 2]
```diff
@@ -24,7 +25,14 @@ export const drawGraphBrush = function(container, store) {
 
         // Only about the main hydrograph when user is done adjusting the time range.
         if (event.sourceEvent.type === 'mouseup' || event.sourceEvent.type === 'touchend') {
-            store.dispatch(Actions.setHydrographXRange(brushRange.map(xScale.invert, xScale)));
+
+            const ts = currentVariablePointsByTsIdSelector('current')(store.getState());
```
mbucknell: I think that this can be simplified considerably. You can get the domain of the brushRange (as was done before to set the hydrograph xrange. You can get the leftmost value of the range by using xScale.range()[0] and then use this to determine the offset. I think the method below will cause issues when the time series has masked data or missing value.

assets/src/scripts/components/hydrograph/graph-brush.spec.js
```diff
@@ -148,7 +148,7 @@ describe ('graph-brush module', () => {
         },
         ui: {
             width: 400,
-            hydrographXRange: undefined
+            hydrographBrushOffset: undefined
```
mbucknell: I think we should move this to timeSeriesState. There will be another brush soon for the dv graph.

assets/src/scripts/components/hydrograph/graph-brush.js
gpetrochenkov-usgs: I went ahead and used x.domain for the same purpose and everything seems to be working.

assets/src/scripts/components/hydrograph/graph-brush.spec.js
gpetrochenkov-usgs: I moved it to time series state.

assets/src/scripts/components/hydrograph/graph-brush.js
```diff
@@ -89,6 +94,6 @@ export const drawGraphBrush = function(container, store) {
 
         }, createStructuredSelector({
             layout: getBrushLayout,
-            isHydrographXRange: (state) => state.ui.hydrographXRange !== undefined
+            isHydrographBrushOffset: (state) => state.ui.hydrographBrushOffset !== undefined
```
mbucknell: we moved the hydrographBrushOffset to timeSeriesState so this needs to change.

assets/src/scripts/store/ui-reducer.js
```diff
@@ -10,28 +10,28 @@ const resizeUI = function(ui, action) {
     };
 };
 
-const setHydrographXRange = function(ui, action) {
+const setHydrographBrushOffset= function(ui, action) {
```
mbucknell: This is now in time-series-state-reducer.js and should be removed from here.

assets/src/scripts/store/ui-reducer.js
```diff
@@ -10,28 +10,28 @@ const resizeUI = function(ui, action) {
...
-            end: action.hydrographXRange[1]
+        hydrographBrushOffset: {
+            start: action.hydrographBrushOffset[0],
+            end: action.hydrographBrushOffset[1]
         }
     };
 };
 
-const clearHydrographXrange = function(ui) {
+const clearHydrographBrushOffset = function(ui) {
```
mbucknell: This is in time-series-state-reducer and should be remove from here.

assets/src/scripts/store/ui-reducer.js
```diff
@@ -10,28 +10,28 @@ const resizeUI = function(ui, action) {
...
+        hydrographBrushOffset: undefined
     };
 };
 
 export const uiReducer = function(ui={}, action) {
     switch(action.type) {
         case 'RESIZE_UI': return resizeUI(ui, action);
-        case 'SET_HYDROGRAPH_X_RANGE': return setHydrographXRange(ui, action);
-        case 'CLEAR_HYDROGRAPH_X_RANGE': return clearHydrographXrange(ui);
+        case 'SET_HYDROGRAPH_BRUSH_OFFSET': return setHydrographBrushOffset(ui, action);
```
mbucknell: These two lines should be removed.

assets/src/scripts/store/ui-reducer.spec.js
```diff
@@ -6,50 +6,15 @@ describe('uiReducer', () => {
         expect(uiReducer({
             width: 50,
             windowWidth: 100,
-            hydrographXRange: undefined
+            hydrographBrushOffset: undefined
```
mbucknell: This should be removed.

assets/src/scripts/store/ui-reducer.spec.js
```diff
@@ -6,50 +6,15 @@ describe('uiReducer', () => {
...
-                start: 150000000,
-                end: 150000100
-            }
-        }, {
-            type: 'CLEAR_HYDROGRAPH_X_RANGE'
-        })).toEqual({
-            width: 50,
-            windowWidth: 100,
-            hydrographXRange: undefined
+            hydrographBrushOffset: undefined
```
mbucknell: This should be removed.

assets/src/scripts/components/hydrograph/scales.js
```diff
@@ -69,13 +69,24 @@ export const createYScale = function (parmCd, extent, size) {
 export const getXScale = memoize((kind, tsKey) => createSelector(
     getLayout(kind),
     getRequestTimeRange(tsKey),
-        state => state.ui.hydrographXRange,
-    (layout, requestTimeRange, hydrographXRange) => {
+        state => state.ui.hydrographBrushOffset,
```
mbucknell: Still using state.ui

CHANGELOG.md
```diff
@@ -11,10 +11,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - All IV time series are now available for drawing on the hydrograph not just those in the last 7 days.
 - Added brush handle fill.
 - Updated to latest wdfn-viz 1.4.0 and implemented the new build guidance.
+- Fixed bug that made compare time series disappear when the brush was adjusted by adding
```
mbucknell: These should be moved to be under Unreleased. I cut a release on Wednesday

CHANGELOG.md
```diff
@@ -11,10 +11,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+offsets to time series state in order to adjust x scale.
 
 ### Added
 - The state of the UI for the hydrograph is now preserved in the hash portion of the url and state 
 can be restored by pasting the url into a browser window.
+- Added upstream basin to NLDI map on the monitoring location pages.
```
mbucknell: Ditto with this one.

jkreft-usgs: The upstream basin is on prod, so it should be in released

mbucknell: I forgot. Yeah this one should stay where it is.

PR #910, opened 2020-03-06
Iow 207 - Allow for different sizes of image to be requested

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Added a width integer parameter. It allows the user to set the width of the returned image in pixels.

After making a pull request
---------------------------

### DIFF
--- a/graph-server/src/index.js
+++ b/graph-server/src/index.js
@@ -84,7 +84,8 @@ app.get(`${PATH_CONTEXT}/monitoring-location/:siteID/`, cache({ttl: CACHE_TIMEOU
         period: req.query.period,
         startDT: req.query.startDT,
         endDT: req.query.endDT,
-        showMLName: req.query.title === 'true'
+        showMLName: req.query.title === 'true',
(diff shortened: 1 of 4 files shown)


graph-server/src/index.js
```diff
@@ -43,12 +43,24 @@ app.get(`${PATH_CONTEXT}/monitoring-location/:siteID/`, cache({ttl: CACHE_TIMEOU
             }
         }
     },
+    width: {
+        in: ['query'],
+        errorMessage: 'width should be an integer',
+        isInt: true,
```
ayan-usgs: Wondering it's easy to add options to this... if so would it be worthwhile to add `{gt: 0}` as a hedge against negative integers? That said, I think the probability of someone doing that is quite low. :thinking:

mbucknell: Yes. I think that's a good idea.

mbucknell: Decided to set min and max to 300 and 1200, respectively. Thanks for the suggestion!

PR #928, opened 2020-03-18
Iow 342 - main axis scaling needs to be more flexible to deal with custom time ranges

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
X Axis scaling is more flexible

Description
-----------
Completely refactored the generateDateTicks function which has been moved to a module in d3-rendering/tick-marks.js and renamed to generateTimeTicks because it is more general now and could potentially be used for other time series graphs. It has a full set of tests. 

After making a pull request
---------------------------


assets/src/scripts/utils.js
```diff
@@ -45,23 +45,6 @@ export const replaceHtmlEntities = function (someString) {
...
 };
 
-/**
- * Calculate the difference in days between two Date objects
- *
- * @param date1
- * @param date2
- * @returns {number}
- */
-export const deltaDays = function (date1, date2) {
```
mbucknell: Removed because we no longe use this utility method and the luxon library has equivalent capability.

assets/src/scripts/d3-rendering/tick-marks.js
```diff
@@ -0,0 +1,173 @@
...
+    let result = {
+        dates: [],
+        format: null
+    };
+
+    if (length.count('hours') <= 4) {
+        // Generates 4 ticks that are on the start of a minute
+        result = {
+            dates: getDefaultTicks(startMillis, endMillis, 'minute', 4, ianaTimeZone),
+            format: formatFnc('MMM dd HH:mm')
```
ayan-usgs: If I recall correctly, D3 uses this function to format the ticks, right?

ayan-usgs: Yep... looks like it.

PR #930, opened 2020-03-19
IOW-309 DV brush

JanellFry-usgs:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


assets/src/scripts/components/dailyValueHydrograph/graph-brush.js
```diff
@@ -0,0 +1,97 @@
+import {brushX, brushSelection} from 'd3-brush';
```
mbucknell: I am going to open a refactor ticket to see if some of this code which is very similar to hydrograph/graph-brush.js can be reusued and moved to d3-rendering. But that does not need to be done right now. Just FYI.

assets/src/scripts/components/dailyValueHydrograph/index.spec.js
```diff
@@ -116,7 +116,7 @@ describe('components/dailyValueHydrograph/index', () => {
         });
 
         window.requestAnimationFrame(() => {
-            expect(testDiv.select('.graph-container').selectAll('.hydrograph-container').size()).toBe(2);
+            expect(testDiv.select('.graph-container').selectAll('.hydrograph-container').size()).toBe(3);
             done();
         });
     });
```
mbucknell: Could you also add a test to make sure the graph-brush is rendered (similar to what you added for the legend)?

assets/src/scripts/components/dailyValueHydrograph/selectors/scales.js
```diff
@@ -8,36 +9,43 @@ import {
...
-    (layout, timeRange) => {
+    state => state.observationsState.dvGraphBrushOffset,
+    (layout, timeRange,dvGraphBrushOffset) => {
         let xScale = scaleLinear();
         if (timeRange) {
             xScale
                 .range([0, layout.width - layout.margin.right])
                 .domain([timeRange.startTime, timeRange.endTime]);
+            if (dvGraphBrushOffset) {
+                xScale.domain([timeRange.startTime+dvGraphBrushOffset.start, timeRange.endTime-dvGraphBrushOffset.end]);
```
mbucknell: Please put leading and trailing spaces around '+' and '-'

assets/src/scripts/components/dailyValueHydrograph/time-series-graph.js
```diff
@@ -59,6 +59,9 @@ const drawDataLines = function (elem, {lines, xScale, yScale, layout}) {
         .attr('y', layout.margin.top)
         .attr('width', layout.width - layout.margin.right)
         .attr('height', layout.height - layout.margin.bottom);
+     if (enableClip) {
```
mbucknell: This is using the graph-clip that was created for the hydrograph. We should keep these components independent. If you need a clip rectangle for the DV graph, you should create one in this module, give it a different id (id's should be unique on a page) and then use that. Do you need it? It doesn't look like you are using it for the DV graph itself. Do you need it for both the main DV graph and the brush? If not remove it, otherwise you should create a clip rectangle for the DV component. I'm a little surprised that this works at all since you have set enableClip for the brush and this rectangle is much taller than what you need there.

JanellFry-usgs: Thanks for catching that, seems to be good now.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added period of record for each parameter in the monitoring location page parameter selection list.
 - Added OpenAPI documentation to the graph-server along with a Swagger UI served at /api/graph-images/api-docs
+- Added Brush to DV graph.
```
mbucknell: This should be added under Unreleased. I did a release this morning.

assets/src/scripts/components/dailyValueHydrograph/time-series-graph.js
mbucknell: I wasn't clear. There isn't a dv-graph-clip rectangle to perform this. I see that the enableClip is not getting set for in drawTimeSeriesGraph when drawing the lines. Do you even need this? What happens if you set enableClip to false in the dv graph-brush.j?

JanellFry-usgs: Oh strange, if I set it to false, they still seem to be clipped.  I swear all the lines were going outside their bounds before this clip was added...hmm

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased](https://github.com/usgs/waterdataui/compare/waterdataui-0.28.0...master)
+- Added Brush to DV graph.
```
mbucknell: Put the ### Added label above this and then it is good to go.

PR #950, opened 2020-04-01
Fix brush offset on dv and uv graphs and fixed the getXScale for brush

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Fix brush offset on dv and uv graphs and fixed the getXScale for brush

Description
-----------
I also streamlines the drawing of the UV data lines and eliminated setting attributes that have no effect.

After making a pull request
---------------------------


assets/src/scripts/components/dailyValueHydrograph/selectors/scales.js
```diff
@@ -9,19 +9,22 @@ import {
...
-            xScale
-                .range([0, layout.width - layout.margin.right])
-                .domain([timeRange.startTime, timeRange.endTime]);
-            if (dvGraphBrushOffset) {
-                xScale.domain([timeRange.startTime + dvGraphBrushOffset.start, timeRange.endTime - dvGraphBrushOffset.end]);
-            }
+        let xScale = scaleLinear()
+            .range([0, layout.width - layout.margin.right]);
+
+        if (kind !== 'BRUSH' && timeRange && dvGraphBrushOffset) {
```
JanellFry-usgs, line 24: This is curious:
if (kind !== 'BRUSH'
shouldn't the kind == Brush here?

mbucknell: The first if clause is for the Main XScale which should be modified with the brush offset when that  and timeRange are available. The next else clause says to use the time range if available to set the scale range. If we have a BRUSH scale it will fail the first if and take this one or the last (if the data has not yet been fetched).

assets/src/scripts/index.spec.js
```diff
@@ -86,3 +86,10 @@ import './web-services/models.spec';
 import './web-services/nldi-data.spec';
 import './web-services/observations-spec';
 import './web-services/statistics-data.spec';
+
+import './networks/web-services/network-data.spec';
```
mbucknell, line 90: @gpetrochenkov-usgs , I added the tests for the networks directory here. It seemed simpler to do this then have a separate karma.conf.js and also the coverage report was not getting generated correctly. Now we can see the coverage (and it's pretty good!).

PR #951, opened 2020-04-02
Added configurable notices that will be added below the Beta notice. …

mbucknell:
…Eventually when the Beta is removed we should make this section only appear if BANNER_NOTICES is not empty.

Before making a pull request
----------------------------

- [X] Make sure all tests run

After making a pull request
---------------------------


mbucknell:
There should be a ticket shortly to add additional notices. We need to add a USGS wide Covid-19 banner.

PR #954, opened 2020-04-03
IOW-433 - Extraneous lines are drawn when using the custom time range more than once.

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Also fixed a bug with setting the custom date from the url.

After making a pull request
---------------------------


assets/src/scripts/store/index.js
```diff
@@ -142,6 +142,15 @@ export const Actions = {
             const state = getState();
             const parmCd = parameterCd;
             const requestKey = getTsRequestKey('current', 'custom', parmCd)(state);
+
+            // We need to resetTimeSeries because the merge function in the addSeriesCollection does not clear out the
```
mbucknell, line 146: When I added the ability to use a custom period I removed this mistakenly thinking we didn't need it. I put in a comment as to why we do this since this is different than the retrieval process we use for standard date ranges.

PR #969, opened 2020-04-10
Iow 391 - Select Daily Value Max from available Time series

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Select Daily Value Max from available Time series

Description
-----------
Fetch the available statistical time series, filter ground water level parameters codes and max statistic code and then fetch that time series and display in the Ground Water levels graph.

I decided to refactor the store for the observations actions and reducers. They are all in the same module. I felt like store/index.js was just too long. I also did not feel like that separate tests for case reducers are useful. Good reducer tests which use the actions are helpful to make sure the actions do what they are suppose to.

I also moved the networks web service out of the networks directory since these is not really page specific and we may want to call the networks service at some time in the future and the monitoring location pages.

After making a pull request
---------------------------


assets/src/scripts/components/dailyValueHydrograph/selectors/time-series-data.js
```diff
@@ -40,6 +43,16 @@ export const getCurrentTimeSeriesPoints = createSelector(
                     'grades'
                 ], zippedStep);
             });
+
+        return result.sort((first, second) => {
```
mbucknell, line 47: I had assumed that the time series data comes in ordered. It doesn't which is why this step is needed.

ayan-usgs: I wonder if this could be addressed in the service in the future.... :thinking:

assets/src/scripts/selectors/observations-selector.js
```diff
@@ -1,42 +1,43 @@
...
-
 /*
- * Return a selector function which returns the current time series id or null if none define
- * @return {Function} - selector function which returns {String}
+ * Selectors to return observations state information or null if missing
  */
-export const getCurrentObservationsTimeSeriesId =
-    (state) => state.observationsState.currentTimeSeriesId ? state.observationsState.currentTimeSeriesId : null;
+export const getCurrentDVTimeSeriesId =
+    (state) => state.observationsState.currentDVTimeSeriesId ? state.observationsState.currentDVTimeSeriesId : null;
```
ayan-usgs, line 8: Is this ternary really necessary or can `||` be used like the lines below?

assets/src/scripts/store/observations.js
```diff
@@ -0,0 +1,184 @@
...
+    return {
+        type: 'ADD_DV_TIME_SERIES',
+        timeSeriesId,
+        data
+    };
+};
+
+/*
+ * Synchronous Redux Action to update the current time series id to be viewed
+ * @param {String} currentDVTimeSeriesId
```
ayan-usgs, line 34: Nit: I think the name of the parameter in here should be `timeSeriesId`.

assets/src/scripts/store/observations.js
```diff
@@ -0,0 +1,184 @@
...
+const setCurrentDVTimeSeriesId = function(timeSeriesId) {
+    return {
+        type: 'SET_CURRENT_DV_TIME_SERIES_ID',
+        timeSeriesId
+    };
+};
+
+/*
+ * Synchronous Redux action to update the graph's cursor offset
+ * @param {Number} dvCursorOffset - difference in epoch time from the cursor position to graph start time
```
ayan-usgs, line 45: Nit: Parameter is actually `cursorOffset`.

assets/src/scripts/web-services/network-data.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const fetchNetworkFeatures = function(networkCd) {
+    const networkQuery = `${networkCd}/items`;
+    return fetchNetworkData(networkQuery)
+        .then((response) => {
+            return response.features ? response.features : [];
+        });
+};
+
+//export const fetchMonitoringLocationAvailableStatisticalData = function(siteId) {}
```
ayan-usgs, line 34: Could this be removed?

assets/src/scripts/components/dailyValueHydrograph/selectors/time-series-data.js
mbucknell: I did ask that question and I was told that for now this is how it will be. I am wondering if I should sort the data before putting it in the Redux store but opted for doing it in the selector. I like the idea of just sticking raw data in the Redux store and having the selectors do the work.

assets/src/scripts/selectors/observations-selector.js
mbucknell: Necessary? No ... I happen to like the ternary syntax better. Is there a best practice around this?

ayan-usgs: I don't think there's a best practice. I was just curious about between line switch away from the ternary.

assets/src/scripts/components/dailyValueHydrograph/selectors/time-series-data.js
jkreft-usgs: yeah, downstream developers are going to be confused without a sort.  Of course, a sort is also really expensive and could potentially slow down a query.  Everything is always tradeoffs.  We might do something like what WQP does, where we allow for sorting as a parameter, but if there is sorting the size of the query is limited.

PR #1013, opened 2020-05-05
Iow 362 - deal with masked data

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Added ability to show DV masked data

Description
-----------
The major changes for this PR are in assets/src/scripts/components/daily-value-hydrograph/selectors/time-series-data.js and legend-data. We needed to be able to identify points and segments that are masked. The points now include the label and class used to decorate each point. This is used when creating the tooltip and legend. 

I also factored out rendering of the svg defs so it can be reused for both the IV and DV graphs.

After making a pull request
---------------------------


mbucknell:
@ssoper-usgs , I'm not sure exactly what you are referring to. The class name is what is passed around. Could you clarify?

PR #1045, opened 2020-05-27
Add flood levels to gage height

gpetrochenkov-usgs:
This work refers to IOW-321: https://internal.cida.usgs.gov/jira/browse/IOW-321

All tests pass.


assets/src/scripts/components/hydrograph/legend.js
```diff
@@ -132,11 +147,14 @@ const legendDisplaySelector = createSelector(
     getCurrentVariableMedianMetadata,
     uniqueClassesSelector('current'),
     uniqueClassesSelector('compare'),
-    (showSeries, medianSeries, currentClasses, compareClasses) => {
+    waterwatchVisible,
+    getWaterwatchFloodLevels,
+    (showSeries, medianSeries, currentClasses, compareClasses, visible, getWW) => {
```
mbucknell, line 152: nit. variable names shouldn't bet a verb. Can you replace getWW with floodLevels (or whatever noun you like)?

assets/src/scripts/components/hydrograph/time-series-graph.js
```diff
@@ -196,6 +255,12 @@ export const drawTimeSeriesGraph = function(elem, store, siteNo, showMLName, sho
             xscale: getMainXScale('current'),
             yscale: getMainYScale,
             seriesPoints: getCurrentVariableMedianStatPoints
+        })))
+       .call(link(store, plotAllFloodLevelPoints, createStructuredSelector({
+            visible: waterwatchVisible,
+            xscale: getBrushXScale('current'),
```
mbucknell, line 261: Shouldn't this be getMainXScale for this graph?

assets/src/scripts/selectors/flood-data-selector.js
```diff
@@ -1,12 +1,17 @@
...
-
+import { getCurrentParmCd } from './time-series-selector';
 
 export const getFloodStages = state => state.floodData.stages || [];
 
 export const getFloodExtent = state => state.floodData.extent || {};
 
 export const getFloodGageHeight = state => state.floodState.gageHeight;
 
+export const getWaterwatchActionStage = state => state.floodState.actionStage;
```
mbucknell, line 10: If we put the stages into a single floodLevel property, I don't think these individual stage ones are necessary

assets/src/scripts/selectors/flood-data-selector.js
```diff
@@ -53,3 +83,16 @@ export const getFloodGageHeightStageIndex= createSelector(
...
+/*
+ * Provides a function which returns the Waterwatch Flood Levels
+ */
+export const getWaterwatchFloodLevels = createSelector(
+    getWaterwatchActionStage,
+    getWaterwatchFloodStage,
+    getWaterwatchModerateFloodStage,
+    getWaterwatchMajorFloodStage,
+    (actionStage, floodStage, moderateFloodStage, majorFloodStage) => {
+        return [actionStage, floodStage, moderateFloodStage, majorFloodStage];
```
mbucknell, line 95: I'm wondering why you returned flood levels as an array rather just return as. The caller of getWaterwatchFloodLevels must then know what the order is. Would it make the code more self documenting if an object was used? I think as I have said elsewhere (probably in the store/flood-inundation module), I think just storing the entire object as is in floodData in a floodLevel property would be more self documenting and make the code simpler.

assets/src/scripts/selectors/flood-data-selector.spec.js
```diff
@@ -78,6 +79,87 @@ describe('flood-data-selector', () => {
...
+                    floodStage: 2,
+                    moderateFloodStage: 3,
+                    majorFloodStage: 4
+                }
+            })).toEqual([1,2,3,4]);
+        });
+    });
+
+       describe('waterwatchVisible', () => {
+        it('Return false if waterwatch flood levels should not be visible', () =>{
```
mbucknell, line 120: I'd like a test for when the parameter code is 00065 but there are no flood levels.

assets/src/scripts/web-services/flood-data.js
```diff
@@ -1,10 +1,11 @@
 import {get} from '../ajax';
 import config from '../config';
 
-
 export const FLOOD_EXTENTS_ENDPOINT = `${config.FIM_GIS_ENDPOINT}floodExtents/MapServer/`;
 export const FLOOD_BREACH_ENDPOINT = `${config.FIM_GIS_ENDPOINT}breach/MapServer/`;
 export const FLOOD_LEVEE_ENDPOINT = `${config.FIM_GIS_ENDPOINT}suppLyrs/MapServer/`;
+const waterwatchUrl = config.WATERWATCH_ENDPOINT;
```
mbucknell, line 7: Please make this all uppercase constants as they are module level constants.

assets/package.json
```diff
@@ -84,6 +84,7 @@
     "autotrack": "2.4.1",
     "core-js": "3.6.5",
     "d3": "5.16.0",
+    "d3-selection": "^1.4.1",
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: I removed this dependency

assets/src/scripts/store/flood-inundation.js
```diff
@@ -65,22 +68,63 @@ const setGageHeight = function(gageHeight) {
...
  */
 export const floodStateReducer = function(floodState={}, action) {
+
     switch(action.type) {
         case 'SET_GAGE_HEIGHT':
             return {
                 ...floodState,
                 gageHeight: action.gageHeight
             };
+        case 'SET_WATERWACH_FLOOD_LEVELS':
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: i moved this to the floodDataReducer.

assets/src/scripts/components/hydrograph/drawing-data.js
```diff
@@ -220,6 +221,46 @@ export const getCurrentVariableMedianStatPoints = createSelector(
...
+    getWaterwatchFloodLevels,
+    getRequestTimeRange('current'),
+    getIanaTimeZone,
+    (floodLevels, timeRange, ianaTimeZone) => {
+        if (!floodLevels || !timeRange) {
+            return [];
+        }
+
+        let datesOfInterest = [];
+        let nextDateTime = DateTime.fromMillis(timeRange.start, {zone: ianaTimeZone});
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: I stripped out the unnecessary bloat here.

assets/src/scripts/store/flood-inundation.js
```diff
@@ -65,22 +68,63 @@ const setGageHeight = function(gageHeight) {
...
+            return {
+                ...floodState,
+                actionStage: action.floodLevels[0] ? parseInt(action.floodLevels[0].action_stage) : null,
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: Added floodLevels to floodData as one property.

assets/src/scripts/store/flood-inundation.js
```diff
@@ -65,22 +68,63 @@ const setGageHeight = function(gageHeight) {
...
+
+/*
+ * Asynchronous Redux action to fetch the Waterwatch flood levels data
+ * @param {String} siteno
+ * @return {Function} which returns a Promise
+ */
+const retrieveWaterwatchData = function(siteno) {
+    return function (dispatch) {
+        const floodLevels = fetchWaterwatchFloodLevels(siteno);
+        return Promise.all([
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: I reduced the function with the caveat that a function had to be returned with a dispatch argument for it to work with the redux middleware.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -101,63 +104,67 @@ export const attachToNode = function (store,
...
-            graphContainer.append('div')
-                .classed('ts-legend-controls-container', true)
-                .call(drawTimeSeriesLegend, store);
 
-            // Add UI interactive elements and the provisional data alert.
-            if (!showOnlyGraph) {
-                nodeElem
-                    .call(drawMethodPicker, store)
-                    .call(drawDateRangeControls, store, siteno);
+    Promise.all([fetchDataPromise, fetchWaterWatchPromise]).then( () => {
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: i removed this portion, I had it in there just in case a waterwatch call lagged but I agree it the returns vs holding up the rest of the processing isn't good.

assets/src/scripts/components/hydrograph/legend.js
gpetrochenkov-usgs: Modified the variable name of getWW to be floodLevels.

assets/src/scripts/selectors/flood-data-selector.js
gpetrochenkov-usgs: Placed all data in to a single flood property.

assets/src/scripts/components/hydrograph/drawing-data.spec.js
```diff
@@ -1029,5 +1035,110 @@ describe('drawingData module', () => {
...
+                        day: 6,
+                        hour: 13,
+                        minute: 45,
+                        second: 0,
+                        zone: 'America/Chicago'
+                    }).valueOf()
+                });
+            });
+
+            it('Return the expected data points', () => {
```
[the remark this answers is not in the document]
gpetrochenkov-usgs: Added a test for no flood data and deleted duplicate.

assets/src/scripts/selectors/flood-data-selector.js
gpetrochenkov-usgs: Changed selector to return as object with properties instead of array.  Sites property of call returns array so not returning exact full object because that would run in to the same array issue as you mentioned.

assets/src/scripts/web-services/flood-data.js
gpetrochenkov-usgs: Changed module level constants to be all upper case.

assets/src/scripts/selectors/flood-data-selector.spec.js
gpetrochenkov-usgs: Added test for correct param code but no flood levels.

assets/src/scripts/components/hydrograph/time-series-graph.js
gpetrochenkov-usgs: Using getBrushXScale updates the dates of the two points with the minimal amount of code in the case of a brush event.  getMainXScale doesn't address brush conditions.  All the tests pass and smoke tests seem OK but let me know if there is anything I'm not considering.

mbucknell: But we are graphing the lines on the main graph. It only appears to work because enableClip is true. I think. At any rate, it seems like if we are using the getMainXScale to scale the time series itself that we should be doing the same thing here. The mainXScale is the one that changes when the brush selection is changed.

gpetrochenkov-usgs: Sounds good, we don't want a fail on enabledClip false.  So here is a picture of the graph when using getXMainScale:
Capture
The issue exists for the median stats as well.  They run off the side of either graph when brushed because the offsets refer to time series sampled at higher resolutions.  Is addressing within the scope of the ticket?  I don't want to assume so.

gpetrochenkov-usgs: Capture2
Sorry I cutoff the brush portion by mistake.  This only happens after a brush event that filters some of the time series.

assets/src/scripts/components/hydrograph/time-series-graph.js
```diff
@@ -86,6 +88,63 @@ const plotAllMedianPoints = function (elem, {visible, xscale, yscale, seriesPoin
...
+ * @param  {Object} elem
+ * @param  {Function} xscale
+ * @param  {Function} yscale
+ * @param  {Number} modulo
+ * @param  {Array} points
+ */
+const plotFloodLevelPoints = function(elem, {xscale, yscale, points, classes}) {
+    const stepFunction = d3Line()
+        .curve(curveStepAfter)
+        .x(function (d) {
```
mbucknell, line 102: So rather than using xscale to translate the date, you draw two points for each line. The y point would be the same, but the x point would just be the extent of the xscale (which you get with xscale.range()). I also believe that you don't need curveStepAfter for a two point line.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ## [Unreleased](https://github.com/usgs/waterdataui/compare/waterdataui-0.32.0...master)
+### Added
```
mbucknell, line 7: we shouldn't have two Added sections. Please add your addition to the existing item under the Added section

assets/src/scripts/components/hydrograph/drawing-data.js
```diff
@@ -220,6 +221,36 @@ export const getCurrentVariableMedianStatPoints = createSelector(
         });
     });
 
+/*
+ * @ return {Array of Arrays of Objects} where the properties are date (universal), and value
+*/
+export const getWaterwatchFloodLevelDataPoints = createSelector(
```
mbucknell, line 227: See the comment on the time-series-graph. I'm thinking this selector might not be needed at all. You would just use the getWaterwatchFloodLevels directly.

assets/src/scripts/components/hydrograph/index.js
```diff
@@ -13,6 +13,7 @@ import {Actions as ivTimeSeriesDataActions} from '../../store/instantaneous-valu
 import {Actions as ivTimeSeriesStateActions} from '../../store/instantaneous-value-time-series-state';
 import {Actions as statisticsDataActions} from '../../store/statistics-data';
 import {Actions as timeZoneActions} from '../../store/time-zone';
+import {Actions as floodDataActions} from '../../store/flood-inundation';
```
mbucknell, line 16: Since the state and data parts of the actions are all in flood-inundation, why don't we call these floodActions. We are inconsistent here. It's called floodDataActions here and floodStateActions in the testing file. Or stick with floodStateActions. But it should be consistent.

assets/src/scripts/components/hydrograph/index.spec.js
```diff
@@ -147,6 +148,20 @@ const TEST_STATE = {
     },
     ui: {
         width: 400
+    },
+    floodData: {
+        floodLevels: {
```
mbucknell, line 153: This is not how we are storing the data now, right? We are just storing a single object and not an object with a sites property with an array value?

assets/src/scripts/components/hydrograph/time-series-graph.js
gpetrochenkov-usgs: I went ahead and used the y value and the value at each end of xscale.domain(), I also removed curveStepAfter.

CHANGELOG.md
gpetrochenkov-usgs: I changed the log to have only one added section.

assets/src/scripts/components/hydrograph/drawing-data.js
gpetrochenkov-usgs: I removed this selector from drawing-data.

assets/src/scripts/components/hydrograph/index.js
gpetrochenkov-usgs: I changed the spec file to be floodData actions to be consistent.

assets/src/scripts/components/hydrograph/index.spec.js
gpetrochenkov-usgs: It isn't how state is storing the data, I went ahead and fixed this file as well as three others.

PR #1082, opened 2020-06-23
Containerized deploy

ayan-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------

### DIFF
--- a/buildJenkinsfile
+++ b/buildJenkinsfile
@@ -52,7 +52,7 @@ pipeline {
         stage('build assets docker image') {
             steps {
                 script {
-                    pipelineUtils.buildDockerImage('wqp_assets', './assets')
+                    pipelineUtils.buildDockerImage('assets', './assets')
                 }
             }
         }
--- a/buildJenkinsfile
+++ b/buildJenkinsfile
@@ -1,0 +1,86 @@
+@Library(value='iow-ecs-pipeline@2.2.0', changelog=false) _
+
+def p
+def deployConfigs
+
+pipeline {
+    agent {
+        node {
+            label 'team:iow'
+        }
+    }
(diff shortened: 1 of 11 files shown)


buildJenkinsfile
```diff
@@ -0,0 +1,86 @@
...
+def deployConfigs
+
+pipeline {
+    agent {
+        node {
+            label 'team:iow'
+        }
+    }
+    parameters {
+        choice(choices: ['snapshots'], description: 'type of build', name: 'BUILD_TYPE')
```
mbucknell, line 13: Good that this is only snapshots. We shouldn't allow releases. I think we could arguably leave out any of the release code below.

pipeline.yml
```diff
@@ -0,0 +1,47 @@
...
+
+# This tells AWS what requests to redirect or route to this container
+urlRules:
+  - type: forward
+    conditions:
+      - "/monitoring-location/*"
+      - "/hydrological-unit/*"
+      - "/networks/*"
+      - "/states/*"
+      - "/components/*"
```
mbucknell, line 42: What is this one for?

buildJenkinsfile
ayan-usgs: Removed the stage where tags are pushed back up to the repo.

pipeline.yml
ayan-usgs: Isn't there an feature-toggled embeddable version of the plots? I guess this could be omitted. :thinking:

mbucknell: There is now that I look closer. Might as well leave in.

PR #1084, opened 2020-06-24
IOW-401 Created a monitoring-location directory. Javascript that is unique to…

mbucknell:
… storing data and rendering that page was moved to this directory. This follows the same pattern as the networks directory.

Before making a pull request
----------------------------

- [X] Make sure all tests run
- [ ] Update the changelog appropriately - Refactoring doesn't really affect how the application is built and used so I think it's okay to skip the changelog for this task.

After making a pull request
---------------------------


mbucknell:
what's the issue with relative imports? They resolve for me when using PyCharm.

gpetrochenkov-usgs:
Just to clarify on the PR (per our discussion) the imports resolved just fine, I mentioned that as a best practice it would be better to use aliases for readability via rollup or an alternative library.

PR #1088, opened 2020-06-29
Iow 442 - Refactor common map code from networks and monitoring-location map components

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Description
-----------
Pulled common creation code for maps and base layers into a module. Also created a legend control by extending Leaflet's control. For code in the new leaflet-rendering directory, I used plain javascript rather than use d3 to select and manipulate the DOM. This will allow us to more easily pick up this code.

After making a pull request
---------------------------


mbucknell:
Regarding globals, they just need to be defined in .eslintrc: https://github.com/usgs/waterdataui/blob/master/assets/.eslintrc#L26

PR #1114, opened 2020-07-27
Radio button additions to parameter selection list (JIRA ticket WDFN-343)

BlakeDraper:
In addition to adding the radio buttons, I also updated some use of the variable name convention `parm` to `param` standing for 'parameter'.  This was done in consultation with @mbucknell.

https://internal.cida.usgs.gov/jira/browse/WDFN-343

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
@@ -117,8 +117,6 @@ export const plotSeriesSelectTable = function (elem,
         .attr('tabindex', 0)
         .attr('role', 'listbox');
 
-
-
     table.append('thead')
         .append('tr')
         .selectAll('th')
@@ -140,11 +138,7 @@ export const plotSeriesSelectTable = function (elem,
         .attr('aria-selected', parm => parm.selected)
         .on('click', function (parm) {
             if (!parm.selected) {
-                // TODO: move this to the redux response. not staying set after store.dispatch.
-                let selectedInput = elem.select('#input-' + parm.variableID);
-                selectedInput.attr('checked', 'true');
                 store.dispatch(Actions.updateIVCurrentVariableAndRetrieveTimeSeries(siteno, parm.variableID));
-               
             }
         })
         .call(tr => {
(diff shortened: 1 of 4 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -117,57 +117,69 @@ export const plotSeriesSelectTable = function (elem,
...
+
+            let paramSelectCol = tr.append('td')
+                .attr('scope', 'row');
+            paramSelectCol.append('input')
+                .attr('type', 'radio')
+                .attr('id', param => `input-${param.variableID}`)
+                .attr('class', 'usa-radio__input param-select-input')
+                .attr('value', param => `${param.variableID}`)
+                .property('checked', param => param.selected)
+                .attr('checked', param => param.selected);
```
mbucknell, line 154: Do you need both the attribute and property? I would think you only need the property.  Can you verify? Also the way to set boolean attributes in D3 is to assign null when you want it to be false. So we typically do this:
.attr('checked', param => param.selected ? true : null)

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -117,57 +117,69 @@ export const plotSeriesSelectTable = function (elem,
...
+        .attr('aria-selected', param => param.selected)
+        .on('click', function (param) {
+            if (!param.selected) {
+                store.dispatch(Actions.updateIVCurrentVariableAndRetrieveTimeSeries(siteno, param.variableID));
+            }
+        })
+        .call(tr => {
+
+            let paramSelectCol = tr.append('td')
+                .attr('scope', 'row');
```
mbucknell, line 147: There should only be one element in a row with 'scope' set to 'row'. This one should probably go since, the radio button itself doesn't really identify the row.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -1,13 +1,13 @@
-import {line} from 'd3-shape';
-import {select} from 'd3-selection';
+import { line } from 'd3-shape';
```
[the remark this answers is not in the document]
BlakeDraper: Yes, VS Code auto-format does this. It's my habit to use. I could update my linter settings there as well.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -117,57 +117,69 @@ export const plotSeriesSelectTable = function (elem,
...
+                store.dispatch(Actions.updateIVCurrentVariableAndRetrieveTimeSeries(siteno, param.variableID));
+            }
+        })
+        .call(tr => {
+
+            let paramSelectCol = tr.append('td')
+                .attr('scope', 'row');
+            paramSelectCol.append('input')
+                .attr('type', 'radio')
+                .attr('id', param => `input-${param.variableID}`)
```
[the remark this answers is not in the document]
BlakeDraper: 👍

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
BlakeDraper: 👍

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
BlakeDraper: Yes, I believe the redundancy is an artifact of some earlier trial and error cycles.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -21,7 +21,7 @@ export const addSparkLine = function(svgSelection, {seriesLineSegments, scales})
         return;
     }
     let spark = line()
-        .x(function(d) {
+        .x(function(d){
```
mbucknell, line 24: nit: we do leave in the space before the bracket. I will likely also add this to our lint rules (space-before-blocks)

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -117,57 +117,67 @@ export const plotSeriesSelectTable = function (elem,
...
 
-
     table.append('thead')
         .append('tr')
-            .selectAll('th')
-            .data(columnHeaders)
-            .enter().append('th')
-                .attr('scope', 'col')
-                .text(d => d);
+        .selectAll('th')
```
mbucknell, line 122: I would like the indenting to be restored here, as indicated before.

BlakeDraper:
@mbucknell updates pushed

mbucknell:
@BlakeDraper , you can go ahead and merge.

PR #1120, opened 2020-08-03
Use the new USWDS Tooltips (JIRA ticket WDFN-402)

mitchas:
I initially missed the tooltips put into the parameters table with .append(), and that method didn't seem to work with the USWDS tooltips. They don't get initialized when they are appended into an element.

I added a custom CSS class to the tooltips added with the appendTooltip function to get them to display without the USWDS initialization. It may not be the best solution - but I wanted to get your thoughts. I'm still digging around in the USWDS code to see if there's a way to initialize them after they are appended.

https://internal.cida.usgs.gov/jira/browse/WDFN-402


assets/src/styles/partials/_tooltip.scss
```diff
@@ -1,63 +1,23 @@
...
-  $tooltip-color: $base-font-color;
-  $tooltip-width: 16em;
-  $tooltip-shadow: 0 2px 2px silver;
-  $tooltip-distance-from-item: 1em;
+// Modified CSS for USWDS Tooltip component
+.usa-tooltip__body {
+    width: auto;
+    white-space: pre-wrap;
+    width: 40vw;
+	max-width: 450px;
```
mbucknell, line 6: Nit: Fix whitespace. Should always indent with two spaces. I hadn't used vw for width before. Nice!

assets/src/styles/partials/_tooltip.scss
```diff
@@ -1,63 +1,23 @@
...
-    margin: 0 auto;
-    width: $tooltip-width;
-    opacity: 0;
-    padding: 1em;
-    text-align: center;
-    transition: all 0.2s ease-in-out;
-    visibility: hidden;
-    z-index: 1000;
+// Rendered USWDS tooltips that aren't initialized by USWDS
+.render-tooltip:hover{
```
mbucknell, line 16: I'm hoping that the work Blake is doing will allow us to not have to put this in here.

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -116,7 +116,10 @@ <h3 class="usa-accordion__heading">
                                 {{ components.DailyValueTimeSeriesComponent(stations[0].site_no) }}
                             </div>
                         </div>
-                    </div>
+					</div>
```
mbucknell, line 119: More whitespace issues. We don't use tabs in our code and html gets indented with 4 spaces.

PR #1134, opened 2020-08-14
WDFN-404: adds network metadata to network pages

BlakeDraper:
JIRA: https://internal.cida.usgs.gov/jira/browse/WDFN-404


BlakeDraper:
Looking into why this test is failing, which causes Travis CI build to fail.

jkreft-usgs:
I think you need to update your requirements.txt (assuming markdown is not part of the standard library)

wdfn-server/waterdata/views.py
```diff
@@ -208,9 +209,14 @@ def networks(network_cd):
     if network_cd:
         collection = network_data
         extent = network_data['extent']['spatial']['bbox'][0]
+        if network_data['properties']['narrative']:
```
mbucknell, line 212: Might consider using python's ternary operator here: https://book.pythontips.com/en/latest/ternary_operators.html

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -46,12 +46,47 @@ <h3 class="usa-alert__heading">Provisional Data Statement</h3>
...
-        <h3>Link Table:</h3>
+        <div id="network-metadata" class="usa-prose">
+            {% if narrative %}
+                <div>{{ narrative|safe }}</div>
+            {% endif %}
+
+            {% if collection.properties.contactInformation.contactName and collection.properties.contactInformation.contactEmail %}
+            <label class="bold-label" for="contact-info">Network Contact</label><br>
+            <span id="contact-info">
+                <span>{{ collection.properties.contactInformation.contactName }}</span>&nbsp;&nbsp;&nbsp;
```
mbucknell, line 61: I'm not a big fan of using &nbsp; to enforce space between elements. Consider setting margin attributes in the style sheet.

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -46,12 +46,47 @@ <h3 class="usa-alert__heading">Provisional Data Statement</h3>
...
+                <span><a href="mailto:{{ collection.properties.contactInformation.contactEmail }}"><i class="fas fa-envelope"></i></a>
+                &nbsp;<a href="mailto:{{ collection.properties.contactInformation.contactEmail }}">{{ collection.properties.contactInformation.contactEmail }}</a></span>
+            </span>
+            {% endif %}
+
+            {% if collection.properties.cooperators %}
+            <div id="network-cooperators">
+                <h3>Cooperators</h3>
+                <div class="grid-container">
```
mbucknell, line 70: I think it's unneeded to use the grid-container class here as it is used to wrap the main content.

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -46,12 +46,47 @@ <h3 class="usa-alert__heading">Provisional Data Statement</h3>
...
+                    <div class="grid-row"> 
+                    {% for cooperator in collection.properties.cooperators %}
+                        <div class="grid-col-12 tablet:grid-col-6 desktop:grid-col-4">
```
mbucknell, line 73: we try to use the Sass mixins rather than the utility classes where possible in our custom markup: https://designsystem.digital.gov/utilities/layout-grid/#sass-mixins. I would like to see utility classes used here.

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -46,12 +46,47 @@ <h3 class="usa-alert__heading">Provisional Data Statement</h3>
...
+                            {% if cooperator.cooperatorLogo == '' %}
+                                <a href="{{ cooperator.cooperatorURL}}">{{ cooperator.cooperatorName }}</a>
```
mbucknell, line 75: We should add the .usa-link class here (see https://designsystem.digital.gov/components/typography/#links - this is kind of buried in the USWDS documentation)

BlakeDraper:
Updated with requested and suggested changes implemented.

PR #1153, opened 2020-08-24
Update filtering functionality for the next round of user research

heymatthenry:
This PR makes a few changes to how the filters work in the prototype monitoring location search:

* The biggest change to functionality is that users have to manually hit a submit button to perform a search, rather than having results update automatically on changes to the filters. (4b14048) This
* All of the filter types of been implemented now, including bounding box (a63d266), time period (36d8064), and parameters (3b35313).
* Baby steps towards testing all of this (f4004f7, 2fffa58). 

Additionally, this PR encompasses some architectural changes to better encapsulate the new WDFN work in its own corner of the codebase:

* Move the component code (including the store) to its own directory (744cc46); and 
* move the filter markup to a template rather than the component macros (ef22dfc).

Aside from functionality like routing that lives outside the template & components of necessity, the only code related to this project remaining outside the `wdfn-home` directory that for a couple of USWDS-related macros that might be useful elsewhere. (a74f523)


assets/src/scripts/web-services/graphql.js
```diff
@@ -10,20 +10,23 @@ query($bBox: String,
       $siteType: [String],
       $startDateLo: String,
       $startDateHi: String) {
-  features(bBox: $bBox,
+  allFeatures(bBox: $bBox,
```
heymatthenry, line 10: This corresponds to changes to the API wrapper made in 10e7743. `allFeatures` gives back the features as well as the total number of sites returned.

assets/src/scripts/wdfn-home/store/wdfn-store.js
```diff
@@ -38,11 +38,11 @@ export const STORE_STRUCTURE = {
 const lookupSiteTypesFullNames = function (siteTypes) {
     // Translate keys into WQP understands
     const dictionary = {
-        groundwater: 'Aggregate groundwater use',
-        surfacewater: 'Aggregate surface-water-use',
+        groundwater: ['Aggregate groundwater use', 'Well', 'Subsurface'],
```
heymatthenry, line 45: This should've been its own commit, but I didn't realize until after pushing. This change just adds more search terms to each umbrella site type grouping.

wdfn-server/waterdata/templates/wdfn.html
```diff
@@ -0,0 +1,173 @@
+{% extends 'base_wdfn.html' %}
```
mbucknell, line 1: I really like how you have implemented this with components for the USWDS elements.

PR #1160, opened 2020-08-27
Add site summaries to search results

heymatthenry:
In anticipation of the next round of user research, this PR adds the following bits of functionality to the prototype:

* Add a sidebar to the search results map that has summary info for each of the monitoring locations returned from the API:
search results

* Show a loading overlay and spinner to give users visual feedback that the search is happening:
loading

* Disable the search button when there are fewer than 2 filters applied (to avoid giant queries)
filter-validation

* Add some non-functional elements to the filter form that are necessary for testing. Wireframe for reference:
Wireframe

* Small changes to make this branch consistent with changes in 9d758c4

### DIFF
--- a/assets/src/scripts/wdfn-home/site-summary/index.js
+++ b/assets/src/scripts/wdfn-home/site-summary/index.js
@@ -1,0 +1,36 @@
+import { select } from 'd3-selection';
+import { link } from '../../lib/d3-redux';
+import { Sites } from '../selectors/wdfn-selector';
+
+const SiteSummaries = (node, store) => {
+    const siteSummaryTmpl = (siteName, siteId) => {
+      return `<li>
+                <h3>${siteName}</h3>
+                <p class="site-id">${siteId}</p>
+              </li>`;
+    };
+
+    const populateSiteSummaries = (node, sites) => {
+      const mapContainer = document.getElementById('wdfn-map-container');
+      const siteSummariesList = document.querySelector('#site-summaries ul');
+
+      if (sites && sites.length > 0) {
+        mapContainer.dataset.summariesVisible = 'true';
+
+        const siteListMarkup = sites.map(s => {
+          return siteSummaryTmpl(s.properties.MonitoringLocationName, 
+            s.properties.MonitoringLocationIdentifier);
+        });
+
+        siteSummariesList.innerHTML = siteListMarkup.join('');
+      }
+    };
+
+    node
+        .call(link(store, populateSiteSummaries, Sites));
+};
+
+export const attachToNode = function(store, node) {
+  select(node)
+    .call(SiteSummaries, store);
+};
--- a/assets/src/scripts/wdfn-home/site-summary/index.js
+++ b/assets/src/scripts/wdfn-home/site-summary/index.js
@@ -20,9 +20,9 @@ const SiteSummaries = (node, store) => {
 
         const siteListMarkup = sites.map(s => {
           return siteSummaryTmpl(
-            s.properties.MonitoringLocationName, 
-            s.properties.MonitoringLocationIdentifier,
-            s.properties.MonitoringLocationTypeName
+            s.properties.monitoringLocationName, 
+            s.properties.monitoringLocationIdentifier,
+            s.properties.monitoringLocationTypeName
           );
         });
 
--- a/assets/src/scripts/wdfn-home/site-summary/index.js
+++ b/assets/src/scripts/wdfn-home/site-summary/index.js
@@ -3,10 +3,11 @@ import { link } from '../../lib/d3-redux';
 import { Sites } from '../selectors/wdfn-selector';
 
 const SiteSummaries = (node, store) => {
-    const siteSummaryTmpl = (siteName, siteId) => {
+    const siteSummaryTmpl = (siteName, siteId, siteType) => {
       return `<li>
                 <h3>${siteName}</h3>
                 <p class="site-id">${siteId}</p>
+                <p class="site-type">${siteType}</p>
               </li>`;
     };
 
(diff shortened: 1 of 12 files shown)


assets/src/scripts/wdfn-home/site-summary/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+      if (sites && sites.length > 0) {
+        mapContainer.dataset.summariesVisible = 'true';
+
+        const siteListMarkup = sites.map(s => {
+          return siteSummaryTmpl(s.properties.MonitoringLocationName, 
+            s.properties.MonitoringLocationIdentifier);
+        });
+
+        siteSummariesList.innerHTML = siteListMarkup.join('');
```
heymatthenry, line 29: Stringifying an array implicitly calls `Array.prototype.join(',')` so explicitly call it here to join with an empty string to avoid rendering extraneous commas.

wdfn-server/waterdata/templates/wdfn.html
```diff
@@ -174,6 +174,14 @@ <h4>
               </div>
             </form>
         </div>
-      {{ USMapComponent(latitude, longitude, zoom) }}
+        <div class="wdfn-component wdfn-map-container"
```
heymatthenry, line 218: the end result

assets/src/scripts/wdfn-home/usmap/index.js
```diff
@@ -53,12 +53,10 @@ const usMap = function(node, {latitude, longitude, zoom}, store) {
...
         features.forEach(f => {
-            if (f.geometry) {
+            if (f.geometry && f.properties) {
                 const marker = L.circle(f.geometry.coordinates.reverse(), {
-                    color: 'red',
-                    fillColor: '#f03',
-                    fillOpacity: 0.2,
-                    radius: 5000
+                    radius: 5000,
+                    className: `site-marker ${f.properties.MonitoringLocationIdentifier}`
```
heymatthenry, line 58: Adding this in here to support another stream of work to allow interactions between the site summaries in the sidebar and the map markers.

assets/src/scripts/wdfn-home/wdfn-param-filters/index.js
```diff
@@ -122,10 +122,32 @@ const WDFNParamFilters = (node, store) => {
     mapContainerEl.dataset.loadingState = loadingState;
   };
 
+  const toggleSearchEnabled = (_, filters) => {
+    let activeFilters = 0;
```
heymatthenry, line 127: To keep users from submitting huge queries, disable the search button when fewer than two filters are applied. This function checks whether at least one value has been specified for at least two filter types (site type, time period, parameter, or geographic)

assets/src/scripts/wdfn-home/store/wdfn-store.js
```diff
@@ -191,11 +191,14 @@ export const retrieveWdfnData = function ({ siteTypes, bBox, parameters, timePer
           variables.startDateLo = timePeriod;
         }
 
+        dispatch(applyLoadingState('loading'));
```
heymatthenry, line 194: Setting this here rather than in the search component makes more sense so it can easily be re-set whenever a search comes back, whether or not there are changed values.

assets/src/scripts/wdfn-home/site-summary/index.js
```diff
@@ -0,0 +1,40 @@
...
+    const siteSummaryTmpl = (siteName, siteId, siteType) => {
+      return `<li>
+                <h3>${siteName}</h3>
+                <p class="site-id">${siteId}</p>
+                <p class="site-type">${siteType}</p>
+              </li>`;
+    };
+
+    const populateSiteSummaries = (node, sites) => {
+      const mapContainer = document.getElementById('wdfn-map-container');
```
mbucknell, line 15: I'm  bothered by passing in node and then not using it to find elements within that node. Eventually, if we carry this to the master branch, I think we should either use D3 for selecting elements or use plain vanilla JS and then subscribe to events using the listen function, https://github.com/usgs/waterdataui/blob/master/assets/src/scripts/lib/d3-redux.js#L73. This function is probably in the wrong place since it doesn't require d3 at all

assets/src/scripts/wdfn-home/wdfn-param-filters/index.js
```diff
@@ -111,10 +112,45 @@ const WDFNParamFilters = (node, store) => {
...
+
+    mapContainerEl.dataset.loadingState = loadingState;
+    if (loadingState == 'loaded') submitBtn.innerText = 'Update results';
+  };
+
+  const toggleSearchEnabled = (_, filters) => {
+    let activeFilters = 0;
+    const filterInstructions = document.querySelector('.filter-instructions');
+
+    const submitBtn = document.getElementById('wdfn-search-submit');
```
mbucknell, line 130: Another thing that I will need to think about going forward if we should think of these components as reusable in which case, using static IDs to select elements seems the wrong thing to do.

assets/src/scripts/wdfn-home/wdfn-param-filters/index.js
```diff
@@ -111,10 +112,45 @@ const WDFNParamFilters = (node, store) => {
...
+
+    const siteTypeFilters = Object.values(filters.siteTypes);
+    if (siteTypeFilters.some(v => v)) activeFilters += 1;
```
mbucknell, line 133: FYI: I just recently put in linting rules to prevent one line if statements in the master branch.

wdfn-server/waterdata/templates/wdfn.html
```diff
@@ -157,17 +170,67 @@ <h4>
                                   </li>
                               </ul>
                           </div>
+                          <div class="filter-param-list grid-col">
+                            <label class="usa-label" for="all-parameters">All USGS Parameters</label>
+                            <div class="usa-combo-box"
+
```
mbucknell, line 176: I would remove all of this whitespace

heymatthenry: Yes, for sure. Done: 065f2cf

assets/src/scripts/wdfn-home/wdfn-param-filters/index.js
heymatthenry: 👍 480a7f5

PR #1164, opened 2020-08-31
Quick bugfix for testing sessions

heymatthenry:
# Remove guard condition preventing queries from running when no site type is selected

One-line change to the filter form. Before implementing the functionality to disable submitting the form when <2 filters were applied, I had a guard in there to require a site type to be selected before running a query just to prevent running queries with no filters. This was just to keep from accidentally hammering the API during development. This validation is handled in the [`toggleSearchEnabled`](https://github.com/usgs/waterdataui/compare/18F...mh-new-wdfn-ui-for-user-research?expand=1#diff-8368f07754b89e3c69bfc225d0305f15L133) method now.


heymatthenry:
Just merging in this small change so it's available for testing sessions later today

PR #1175, opened 2020-09-03
Wdfn 368 make brush less confusing

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-368 Make Brush Less Confusing
-----------
Adds hint text and custom 'handles' to the brush area to make it more obvious to users what to do

Description
-----------
Image of new brush handles and text
image

A more mobile view
image

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -12,19 +12,32 @@ import {getBrushLayout} from './selectors/layout';
...
+    let customHandle;
+    let layoutHeight;
 
     const brushed = function() {
+        // if the user clicks a point in the brush area without making an actual selection, remove the custom handles
+        if (event.selection == null) {
+            customHandle.attr('display', 'none');            
+        }
+        customHandle.attr('transform', function(d, index) {
+            const yPositionForCustomHandle = mediaQuery(config.USWDS_LARGE_SCREEN) ? layoutHeight / 3.3 * -1 : layoutHeight / 2.5 * -1;
```
mbucknell, line 28: Assuming that 3.3 and 2.5 are just "magic" constants. Rather than multipling by -1 can we just apply the negative sign to layoutHeight?

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -12,19 +12,32 @@ import {getBrushLayout} from './selectors/layout';
...
+            return 'translate(' + [event.selection[index], yPositionForCustomHandle] + ')';
```
mbucknell, line 29: Can we use a string literal here? I know that sometimes this doesn't work (I have not been able to determine when that is).

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -41,11 +54,19 @@ export const drawGraphBrush = function(container, store) {
...
                 elem.attr('viewBox', `0 0 ${layout.width + layout.margin.left + layout.margin.right} ${layout.height + layout.margin.bottom + layout.margin.top}`);
             }, getBrushLayout
-            ))
+        ))
         .call(svg => {
+            svg.append('text')
+                .classed('brush-text-hint', true)
+                .text('drag handles to change timeframe')
+                .attr('text-anchor', 'middle')
+                .attr('font-size', 'smaller')
```
mbucknell, line 63: I would like the font-size to be specified in the style sheet, preferably using a USWDS design token.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -62,23 +83,38 @@ export const drawGraphBrush = function(container, store) {
...
             svg.select('.brush').remove();
 
             const group = svg.append('g').attr('class', 'brush')
-                .attr('transform', `translate(${layout.margin.left},${layout.margin.top})`);
+                .attr('transform', `translate(${layout.margin.left}, ${layout.margin.top})`);
 
+            graphBrush.handleSize([1]); // make default handle 1px wide
             graphBrush.extent([[0, 0], [layout.width - layout.margin.right, layout.height - layout.margin.bottom - layout.margin.top]]);
 
+            const brushResizePath = function(d) {
```
mbucknell, line 99: A short comment here would be good to explain what this does.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -62,23 +83,38 @@ export const drawGraphBrush = function(container, store) {
...
+                let e = +(d.type == 'e'),
```
mbucknell, line 100: Not crazy about this method of creating a numeric. Also not big on single character variable names (x and y are okay because those are standard). Consider at least changing this to e = d.type === 'e' ? 1 : 0.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -62,23 +83,38 @@ export const drawGraphBrush = function(container, store) {
...
+                    x = e ? 1 : -1,
+                    y = layoutHeight / 2;
+                return 'M' + (.5 * x) + ',' + y + 'A6,6 0 0 ' + e + ' ' + (6.5 * x) + ',' + (y + 6) + 'V' + (2 * y - 6) + 'A6,6 0 0 ' + e + ' ' + (.5 * x) + ',' + (2 * y) + 'Z' + 'M' + (2.5 * x) + ',' + (y + 8) + 'V' + (2 * y - 8) + 'M' + (4.5 * x) + ',' + (y + 8) + 'V' + (2 * y - 8);
```
mbucknell, line 103: Consider using a string literal here. This is nearly impossible to parse.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -62,23 +83,38 @@ export const drawGraphBrush = function(container, store) {
...
+            };
+            customHandle = group.selectAll('.handle--custom')
+                .data([{type: 'w'}, {type: 'e'}])
+                .enter().append('path')
+                .attr('class', 'handle--custom')
+                .attr('stroke', '#000')
```
mbucknell, line 109: Consider putting the stroke and cursor attributes in the style sheets

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -90,7 +126,7 @@ export const drawGraphBrush = function(container, store) {
                 selection = xScale.range();
             }
             if (selection[1] - selection[0] > 0) {
-                graphBrush.move(group, selection);
+                group.call(graphBrush.move, selection);
```
mbucknell, line 129: This does the same thing, FWIW. I'm not sure there is a preference other than my own.

mbucknell (changes requested):
There are some comments on things that you should consider, but I would like the items noted that could be moved to the stylesheets to be moved.

Also, there is a second place on the pages where we use the brush, the dailyValueHydrograph. See https://webvadevvs04.er.usgs.gov/wsgi/waterdataui/monitoring-location/354133082042203/#parameterCode=72019 . I have not been able to refactor this in a satisfactory way so that changes would only need to be made in one place, but if you see a way to do this, please take a crack. But it's not necessary for this ticket. But we would like both brushed to looks the same.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: lol. Why didn't I think of that?? Also will add the magic numbers as constants with names that might make their function clearer.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: Will do.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: Sure can.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: Yup, no preference here either. I was just experimenting and forgot switch it back.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: Makes sense will do.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: This took some fiddling, but I think you're right a string literal makes gibberish into slightly less gibberish :)

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: Will do.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
abriggs-usgs: I'll do my best.

abriggs-usgs:
I updated the DVHydrograph with the custom brush handles. Since the DVHydrograph has dates on the x-axis, I didn't feel like placing the hint under the brush area worked as well as it did in the regular hydrograph without dates, so I positioned things just a bit differently. I like this layout (hint above) when the dates are present on the x-axis, but I prefer the 'hint' below the x-axis when there are no dates. I can also see some value in having the hint positioned similarly in on both graphs. In that case, I would favor moving the hint to the top of both as there is too much information clustered on the bottom when both the hint and the dates are present. Here's how it looks. 
desktop --
image
mobile --
image

mbucknell:
So the Hydrograph brush does have an x axis, it just has a height of zero. I'm not sure when this regression happened but it is this way all the way out on production. I think we need to have the x axis back so I will open a bug ticket. In the meantime I think both help messages should appear above the brush. We want these elements to look and feel as similar as possible.

abriggs-usgs:
Sounds good. Will do.

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.js
```diff
@@ -73,14 +93,54 @@ export const drawGraphBrush = function(container, store) {
...
+            graphBrush.handleSize([1]); // make default handle 1px wide
             graphBrush.extent([[0, 0], [layout.width - layout.margin.right, layout.height - layout.margin.bottom - layout.margin.top]]);
 
+            /* Draws the custom brush handle using an SVG path. The path is drawn twice, once for the handle
+            * on the left hand side, which in d3 brush terms is referred to as 'east' (data type 'e'), and then
+            * inverted for the right hand custom handle. Here 'east' will be a value of either 1 or 0 (in effect, making
+            * it a boolean value of 'east' or 'not east' */
+            const brushResizePath = function(d) {
+                let east = d.type === 'e' ? 1:0,
+                    x = east ? 1 : -1,
```
mbucknell, line 105: I need to see if there is a linting rule for this. IOW's convention is to have each variable declared separately. Also all three of these variables are constants therefore should use 'const' rather than 'let'.

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.js
```diff
@@ -73,14 +93,54 @@ export const drawGraphBrush = function(container, store) {
...
+            // Add a class so the default handles can have styling that won't conflict with the slider handle
+            svg.selectAll('.handle').classed('standard-brush-handle', true);
+
+            svg.select('.brush-text-hint').remove();
+            // Add the hint text after the brush is created so that the words sit on top of the brush area
+            svg.call(svg => {
+                svg.append('text')
+                    .classed('brush-text-hint', true)
+                    .text('drag handles to change timeframe')
+                    .call(link(store,(elem, layout) => elem.attr('transform', `translate(${layout.width / 2 + layout.margin.left}, 10)`),
```
mbucknell, line 140: sorry that I didn't catch this before. Nesting calls to "link" as is done here can cause very strange bugs to appear. I would restructure this to draw the text hint outside the link call on line 84. It only needs to be redrawn when the layout changes so I think it will just be a matter of drawing it after this "link" call. And if that doesn't work to draw it on top, I suggest styling it with a z-index.

assets/src/scripts/monitoring-location/components/hydrograph/graph-brush.js
```diff
@@ -62,23 +79,64 @@ export const drawGraphBrush = function(container, store) {
...
+                .attr('class', 'handle--custom')
+                .attr('d', brushResizePath);
+
             // Creates the brush
             group.call(graphBrush);
 
-            // Fill & round corners of brush handles
-            svg.selectAll('.handle').classed('brush-handle-fill', true)
-                .attr('rx', 15).attr('ry', 15);
+            svg.select('.brush-text-hint').remove();
```
mbucknell, line 127: See comment above in dailyValueHydrgoraph

mbucknell (changes requested):
Sorry that I didn't catch the nested "link" calls. That will need to be fixed.

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.js
abriggs-usgs: Ah, now that you mention it, I seem to recall a distant memory of that rule. :) Fixed.

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.js
abriggs-usgs: No problem, I do appreciate your detailed code reviews--hugely helpful. 

Anyway, after going down one dead end after another chasing the exact right combination of position and layer order and finding that SVGs don't like z-index and that the brush area rectangle would redraw over my nicely positioned words when the handles were moved, I finally comprised and moved the text up so that it doesn't overlap at all, which was certainly the easier solution, if not the better looking one as well.
image

mbucknell (approved):

PR #1190, opened 2020-09-16
Wdfn 426 brush axis missing tick marks

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN 426 Brush Axis Missing Tick Marks
-----------
Returns tick marks to brush axis

Description
-----------
The bug was caused by arguments being passed in the incorrect order. This caused the time zone received by the function to be the period code, which is, as might be imagined, not going to turn out well. Switching the order cleared up the issue. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.spec.js
+++ b/assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.spec.js
@@ -70,6 +70,7 @@ describe ('monitoring-location/components/daily-value-hydrograph/graph-brush mod
             expect(div.select('.brush-text-hint').size()).toBe(1);
             expect(div.selectAll('.handle').size()).toBe(2);
             expect(div.selectAll('.handle--custom').size()).toBe(2);
+            expect(div.selectAll('.tick').size()).toBe(5);
         });
 
         it('Should create a DV line, and an x-axis', () => {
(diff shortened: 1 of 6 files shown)


assets/src/scripts/monitoring-location/components/daily-value-hydrograph/graph-brush.spec.js
```diff
@@ -70,6 +70,7 @@ describe ('monitoring-location/components/daily-value-hydrograph/graph-brush mod
             expect(div.select('.brush-text-hint').size()).toBe(1);
             expect(div.selectAll('.handle').size()).toBe(2);
             expect(div.selectAll('.handle--custom').size()).toBe(2);
+            expect(div.selectAll('.tick').size()).toBe(5);
```
mbucknell, line 73: This is great! Adding tests for the regression condition!

assets/src/scripts/monitoring-location/components/hydrograph/selectors/axes.js
```diff
@@ -77,7 +78,7 @@ export const getBrushXAxis = createSelector(
     getBrushXScale('current'),
     getTsTimeZone,
     getCurrentDateRangeKind,
-    (xScale, ianaTimeZone, period) => createXAxis(xScale, period, ianaTimeZone)
+    (xScale, ianaTimeZone, period) => createXAxis(xScale, ianaTimeZone, period)
```
mbucknell, line 81: Looks like we only need the time zone and not the period.

mbucknell (approved):

assets/src/scripts/monitoring-location/components/hydrograph/selectors/axes.js
abriggs-usgs: Oh, yup, duh. I'll fix it!

PR #1193, opened 2020-09-22
Wdfn 352 select arbitrary number of days

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN 352 Select Arbitrary Number of Days
-----------
These changes allow the user to select a up to 99,999 days  of data before to today to display on the graph

Description
-----------
This adds an additional set of radio buttons that only appear when the 'Custom' radio button is selected. These buttons allow users to choose from two input forms 
image
image
When the user selects the time period, it appears in the URL
image
The URL with the period parameter can also entered manually and the custom 'days before today' will open and self populate.

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -188,17 +338,39 @@ export const drawDateRangeControls = function(elem, store, siteno) {
...
             const selectedVal = selected.attr('value');
+
+            // Remove any values stored in the form, because they may not match what is shown in the graph until the submit button is pushed
+            store.dispatch(ivTimeSeriesStateActions.setUserInputNumberOfDays(''));
+            store.dispatch(ivTimeSeriesStateActions.setCustomIVTimeRange(null));
+
             if (selectedVal === 'custom') {
-                customDateContainer.attr('hidden', null);
+                // Remove any values stored in the form, because they may not match what is shown in the graph until the submit button is pushed
+                store.dispatch(ivTimeSeriesStateActions.setUserInputNumberOfDays(''));
```
mbucknell, line 351: This code (that dispatches the actions) can be pulled out of the if clause because it is always executed (it is repeated in the else section)

assets/src/scripts/monitoring-location/components/hydrograph/index.js
```diff
@@ -97,7 +97,8 @@ export const attachToNode = function (store,
             .then(() => {
                 // Fetch any extended data needed to set initial state
                 const currentParamCode = parameterCode ? parameterCode : getCurrentParmCd(store.getState());
-                if (period === 'P30D' || period === 'P1Y') {
+
+                if (period && period.substr(0,1) === 'P' && period !== 'P7D') {
```
mbucknell, line 101: In general, we don't do checking on the hash parameters. Maybe we should but we don't anywhere else.

assets/src/scripts/monitoring-location/store/instantaneous-value-time-series-data.js
```diff
@@ -257,7 +259,14 @@ const retrieveExtendedIVTimeSeries = function(siteno, period, paramCd=null) {
         const state = getState();
         const thisParamCd = paramCd ? paramCd : getCurrentParmCd(state);
         const tsRequestKey = getTsRequestKey ('current', period, thisParamCd)(state);
+        const currentUserInputCustomTimeRangeSelectionButton = period !== 'P7D' && period !== 'P30D' && period !== 'P1Y' ? 'custom' : period;
```
mbucknell, line 262: I wonder if a small helper function for the condition part of the ternary operator would be a good thing. You could also use it in line 265.

assets/src/scripts/utils.js
```diff
@@ -146,25 +146,25 @@ export const mediaQuery = function (minWidth) {
...
- * @param {String} period -- ISO duration for date range of the time series
+ * @param {String} period -- ISO duration for date range of the time series - which has the form of something like
+ * P7D in which the 'P' stands for 'period', the '7' is units in the period, and 'D' is the type of unit
+ * in this case 'Days.' There are only two unit types here, 'Days' or 'Years'
  * @param {Number} endTime -- the end time as universal time
  * @param {String} ianaTimeZone -- Internet Assigned Numbers Authority designation for a time zone
  * @returns {Number} the start time as universal time
  */
-export const calcStartTime = function (period, endTime, ianaTimeZone) {
+export const calcStartTime = function(period, endTime, ianaTimeZone) {
```
mbucknell, line 156: So we only did this before so that we could fetch the "compare" data. I don't think you need to do this for the custom days from today but rather just use the period parameter call when fetching the data. See previous comment in the date-controls.js module.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -31,32 +34,177 @@ export const drawDateRangeControls = function(elem, store, siteno) {
...
+            // Validate user input for things not a number and blank entries
+            if (isNaN(userSpecifiedNumberOfDays) || userSpecifiedNumberOfDays.length === 0) {
+                customDaysBeforeTodayAlertBody.selectAll('p').remove();
+                customDaysBeforeTodayAlertBody.append('p')
+                    .text('Entry must be a number.');
+                customDaysBeforeTodayValidationContainer.attr('hidden', null);
+            } else {
+                customDaysBeforeTodayValidationContainer.attr('hidden', true);
+                store.dispatch(ivTimeSeriesStateActions.setUserInputNumberOfDays(userSpecifiedNumberOfDays));
+                store.dispatch(ivTimeSeriesDataActions.retrieveExtendedIVTimeSeries(
```
mbucknell, line 196: I don't think you should be using retrieveExtendedIVTimeSeries. This will also retrieve the compare data (one year in the future) and we are not allowing that for custom time period or date ranges. I think you should be using retrieveCustomTimePeriodIVTimeSeries.

mbucknell (changes requested):
Overall just a fantastistic job. My biggest worry is that we should not be using retrieveExtendedTimeSeries for the custom days before today. This also retrieves the "previous year's data" and that could be a lot of data. We do not allow compare for custom dates.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -10,13 +10,16 @@ import components from '../../../../../node_modules/uswds/src/js/components';
...
+    getUserInputTimeRangeSelectionButton,
+    getUserInputCustomTimeRangeSelectionButton,
+    getUserInputNumberOfDays,
     getCustomTimeRange} from '../../selectors/time-series-selector';
 import {getIanaTimeZone} from '../../selectors/time-zone-selector';
 import {Actions as ivTimeSeriesDataActions} from '../../store/instantaneous-value-time-series-data';
 import {Actions as ivTimeSeriesStateActions} from '../../store/instantaneous-value-time-series-state';
 
 export const drawDateRangeControls = function(elem, store, siteno) {
+    const MAX_NUMBER_OF_DAYS_USERS_CAN_INPUT = 5;
```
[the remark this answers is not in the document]
abriggs-usgs: Good call. Names are hard :)

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -31,32 +34,177 @@ export const drawDateRangeControls = function(elem, store, siteno) {
         period: 'custom',
         ariaExpanded: false
     }];
+    const CUSTOM_TIMEFRAME_RADIO_BUTTON_DETAILS = [
+        {
+            id: 'custom-input-days-before-today',
+            value: 'days',
+            text: 'days before today',
```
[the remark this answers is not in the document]
abriggs-usgs: Good catch. Fixed

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -31,32 +34,177 @@ export const drawDateRangeControls = function(elem, store, siteno) {
...
+            checked: true,
+            ariaExpanded: true
```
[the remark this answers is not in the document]
abriggs-usgs: I actually went round and round with this one. I didn't know the best route, since the 'days before today' is hidden, but it open when hidden. So I changed it back and forth a couple of times. But I'm happy to go with false.

assets/src/scripts/monitoring-location/store/instantaneous-value-time-series-data.js
abriggs-usgs: Oh, neat idea. Will do.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Very true. Fixed

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Finally got this working. Thanks for the help!

assets/src/scripts/monitoring-location/components/hydrograph/index.js
abriggs-usgs: I think I understand what you mean here, but I might be confused. Do you mean checking that 'period' is not null or undefined? I didn't do that first and ended up with a set of tests failing.

mbucknell: So the test is testing to make sure nothing other than P30D or P1Y is passed. Now it needs to test PxxD or P1Y. So this check needs to change and the tests need to change

abriggs-usgs: Well, I see that I was missing out on excluding years other than P1Y, but I think that is fixed now.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -31,32 +34,177 @@ export const drawDateRangeControls = function(elem, store, siteno) {
...
+        .attr('class', 'usa-radio__label')
+        .attr('for', (d) => `${d.value}-input`)
+        .text((d) => d.text);
+    listItemForCustomSelectRadioButtons.call(link(store, (elem, userInputCustomTimeRangeSelectionButton) => {
+       elem.select(`#${userInputCustomTimeRangeSelectionButton}`).property('checked', true);
+    }, getUserInputCustomTimeRangeSelectionButton));
 
-    const dateRangePicker = customDateContainer.append('div')
+    // Add controls for selecting time in days from today
+    const numberOfDaysSelection = containerCustomDaysBeforeToday.append('div')
```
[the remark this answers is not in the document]
abriggs-usgs: I was thinking that this wasn't the easiest code to read. Breaking it up helps a bit.

abriggs-usgs:
I ended up backtracking a bit and created a new branch, so I'll close this pull request for the new one, https://github.com/usgs/waterdataui/pull/1213 .

PR #1206, opened 2020-09-29
Add File Types for Android Icon Set to Gulp Processing

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Add File Types for Android Icon Set to Gulp Processing
-----------
Hopefully this will fix the failed deploy to dev

Description
-----------
Only certain file types were being processed by Gulp for hashing. Since the new icon set contained some Android files of type 'webmanifest' and 'xml', I added those to the list of files for hashing. Hopefully this is the best choice. At first I created a new filter for non-hashed assets, but that seemed like more than what was needed.

After making a pull request
---------------------------


mbucknell (approved):
I think it's fine to add these. The purpose of hashing the assets is so browsers reload that static resource. Since these are also static resources, I think it is appropriate to hash them

PR #1208, opened 2020-09-30
Wdfn 420 cookie for covid message

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN 420 Cookie for COVID Message
-----------
Sets a cookie so that the COVID message only appears once a month

Description
-----------
As far as the cookie goes, this works. But I have been struggling to find a way to make a unit test for this. So, I am wondering if perhaps there is a better way to structure the code to make for easier testing. Since this is my first time working with cookies, I am open to thoughts on how to structure this and I thought it might be best to make sure the actual cookie setting made sense before I invested too much more time in figuring out the unit tests.

After making a pull request
---------------------------


mbucknell (commented):
It looks fine to me.

mbucknell (approved):

PR #1220, opened 2020-10-19
Iow 541 v3 rollup alias plugin

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Add Rollup/plugin-alias
-----------
Adds the a plugin to Rollup to let use reference module imports by name rather than importing then with complex relative paths

Description
-----------
Same as noted above 

After making a pull request
---------------------------


assets/package.json
```diff
@@ -81,6 +81,7 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "5.14.0",
+    "@rollup/plugin-alias": "^3.1.1",
```
mbucknell, line 84: This should be in devDependencies since it is used during the build process.

mbucknell (approved):
See comment about moving the rollup plugin alias dependency. Otherwise it looks great!

assets/package.json
abriggs-usgs: Easy fix. I'll take care of it.

PR #1241, opened 2020-11-05
Wdfn 275 use png for hydrograph for ie11

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-275 Use PNG for Hydrograph for Internet Explorer 11
-----------
Swap the interactive hydrograph for a static version

Description
-----------
I added code that will in theory swap out the hydrograph. The code works perfectly to swap the images on Chrome and Firefox and has been written with the plain ole JavaScript that Internet Explorer prefers. 

However, it doesn't work.

When tested in Browserstack's implementation of Internet Explorer 11, the code hits an error right at the as it begins processing the JavaScript side of the application and the code execution stops prior to reaching the new code. It seems that this existing error would need resolution before the new code can function. 

I did some preliminary investigation and attempted to add the Core-js Object.assign polyfill, but this did not produce results and it seems the issue will require deeper investigation. 

Update - changed direction on this and implemented the hydrograph/static image switch on the Flask/Jinja2 side of things. The new implementation successfully produces a static image on the page. 

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/index.js
```diff
@@ -141,18 +152,28 @@ export const attachToNode = function(store,
...
-                    .call(drawGraphBrush, store);
-            }
-
-            graphContainer.append('div')
-                .classed('ts-legend-controls-container', true)
-                .call(drawTimeSeriesLegend, store);
+                .call(link(store, controlDisplay, hasAnyTimeSeries));
 
+            if (window.navigator.userAgent.indexOf('MSIE') !== -1) {
+                graphContainer.call(link(store, getStaticGraph, createStructuredSelector({
```
mbucknell, line 158: If D3 doesn't work in IE11, this won't work either. Right now, I believe on IE11, you will not be able to change the parameterCode or anything else because all of that code is rendered using D3 and doesn't appear. 

I think we may trying to be too smart here. What if we just add a snippet of javascript to https://github.com/usgs/waterdataui/blob/master/wdfn-server/waterdata/templates/base_plain.html to detect the browser and then add the img tag with the png.

mbucknell (commented):
See comment for another suggestion on how to implement this.

assets/src/scripts/monitoring-location/components/hydrograph/index.js
abriggs-usgs: That is an easy way to do it, and I implemented a version of that on day one. But then when I tested it, I realized that all the page function is lost and people can only get from page to page by an existing link or by direct entry of the URL. Is that an acceptable compromise? I guess, it maybe the only way to get this to work.

mbucknell: But as it stands now, there is nothing on the page except the summary of all available data and location metadata. They can't change anything anyway. At least this way they get a hydrograph.

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -85,14 +85,16 @@ <h1>{{ stations[0].station_nm }}</h1>
                             var splitUrl = urlString.split('#')
                             var siteNumber = splitUrl[0].split('/monitoring-location/')[1].slice(0, -1);
                             var paramString = splitUrl[1];
+
+                            if (typeof(paramString) === "undefined") {
+                                paramString = 'parameterCode=00060'
```
mbucknell, line 90: The default parameter is not always 00060, specifically for sites that don't have Discharge. You should be able to get the "default parameter" by looking at the template variable parm-grp_summary: https://github.com/usgs/waterdataui/blob/master/wdfn-server/waterdata/views.py#L126

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -162,22 +164,22 @@ <h2 class="usa-accordion__heading">
                         <div id="ts-a1" class="usa-accordion__content">
                             <table class="usa-table" id="site-data-summary">
                                 <thead>
-                                    <tr>
-                                        <th scope="col">USGS Parameter Group</th>
-                                        <th scope="col">Data Types</th>
-                                        <th scope="col">Start Date</th>
-                                        <th scope="col">End Date</th>
-                                    </tr>
+                                <tr>
```
mbucknell, line 167: Nit: This should be indented.

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -194,33 +196,33 @@ <h2 class="usa-accordion__heading">
                     <div id="ts-a2" class="usa-accordion__content">
                         <table class="usa-table" id="site-summary" summary="Descriptive metadata about this monitoring location">
                             <thead>
-                                <tr>
-                                    <th scope="col">Metadata Element</th>
-                                    <th scope="col">Location Metadata</th>
-                                    <th scope="col">Metadata Code</th>
-                                </tr>
+                            <tr>
```
mbucknell, line 199: More indenting

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -194,33 +196,33 @@ <h2 class="usa-accordion__heading">
...
-                                                {{ value.name }}
-                                            {% endif %}
-                                            {% if value.desc %}
-                                                {{ components.QuestionTooltip(value.code, value.desc, False) }}
-                                            {% endif %}
-                                        </td>
-                                        <td>{% if value.code != value.name %}{{ value.code }}{% else %}<sub>n/a</sub>{% endif %}</td>
-                                    </tr>
-                                {% endfor %}
+                            {% for key, value in location_with_values.items() %}
```
mbucknell, line 206: More indenting

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -78,48 +78,70 @@ <h1>{{ stations[0].station_nm }}</h1>
                     <p id="site-description">{{ components.Description(stations[0].site_no, location_with_values, parm_grp_summary) }}</p>
                 </div>
 
-                {{ components.TimeSeriesComponent(stations[0].site_no, stations[0].dec_lat_va, stations[0].dec_long_va) }}
+                {% if request.user_agent.browser == 'msie' %}
+                    {% block javascript %}
```
mbucknell, line 82: I don't think a block is necessary here. Unless this block will be overridden somewhere else this should be removed.

wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -78,48 +78,70 @@ <h1>{{ stations[0].station_nm }}</h1>
...
+                        <script>
+                            var urlString = window.location.href;
```
mbucknell, line 84: I am a little worried about putting javascript here that selects a node. It might be better to wrap this in a document.addEventListener('DOMContentLoaded', function(event) {......})

mbucknell (changes requested):
Several nits on indenting, but the two import comments are that the default parameter is not always 00060 and wrapping the javascript in a DOMContentLoaded event handler.

wdfn-server/waterdata/templates/monitoring_location.html
abriggs-usgs: Oh, sorry, I cut an replaced the page and the the IDE decided that this was the way to go. lol. I'll change it.

wdfn-server/waterdata/templates/monitoring_location.html
abriggs-usgs: Makes sense--will do.

mbucknell (approved):

PR #1264, opened 2020-11-17
Wdfn 445 add link to waterservices for rdb download

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN 445 Add Link to WaterServices for RDB Download
-----------
Add links so that users can download data related to the current graph

Description
-----------
This adds an accordion with four links, two of which are always available and the remaining two are dependent on if the user has selected the 'compare to last year' or 'display median' check boxes. The links are dynamically created so that the downloaded data will correspond to the selections on the hydrograph.

Image with default selections
image

Image with Median and Compare selected
image

Update - after review I rearranged things, fixed and issue where the ' extended compare' urls stored in the state were not getting the parameter code and added tool tips.
image

After making a pull request
---------------------------


mbucknell (changes requested):
I really think we should be trying to grab the URLs for the data from the state. Let's chat if you have issues with this.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -0,0 +1,69 @@
...
+
+import config from 'ui/config.js';
+
+/*
+* Converts a numerical time representation to a calendar date format
+* @param {Object} customTimeRange - A numerical time indicator than can be converted to date format
+* @param {String} ianaTimeZone - A geographical reference to the user's timezone
+* @return {Object} Contains the start and end calendar dates
+*/
+export const convertTimeToDate = function(customTimeRange, ianaTimeZone) {
```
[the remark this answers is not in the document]
abriggs-usgs: Changed things around so this test is not needed.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -0,0 +1,69 @@
...
+* Assembles the href/URL needed to contact WaterServices and return data related to the currently showing hydrograph
+* @param {String} currentIVDateRange - The string will be in the form of P{number of days}D (like P7D) or P1Y or 'custom'
+* If the value is 'custom', it means that the user selected the timespan in calendar days
+* @param {Object} customTimeRange - A numerical time indicator than can be converted to date format
+* This will only have a value if the currentIVDateRange is custom
+* @param {String} ianaTimeZone - A geographical reference to the user's timezone
+* @param {String} parameterCode - the five digit code for the current hydrograph parameter
+* @ return {String} a URL formatted to return data from WaterServices that matches the currently displayed hydrograph
+*/
+export const createStationDataDownloadURLForWaterServices = function(currentIVDateRange, customTimeRange, ianaTimeZone, parameterCode, siteno) {
```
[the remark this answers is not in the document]
abriggs-usgs: Shortened!

wdfn-server/config.py
```diff
@@ -55,6 +55,11 @@
 # WaterAlert
 WATERALERT_SUBSCRIPTION = 'https://water.usgs.gov/wateralert/subscribe2'
 
+#WaterServices
+WATER_SERVICES_IV = 'https://waterservices.usgs.gov/nwis/iv'
```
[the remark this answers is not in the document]
abriggs-usgs: Fixed

assets/src/scripts/monitoring-location/components/hydrograph/index.js
```diff
@@ -164,6 +167,70 @@ export const attachToNode = function(store,
 
                 nodeElem.select('.ts-legend-controls-container')
                     .call(drawGraphControls, store);
+
+                // Construct and add the hrefs needed so users can download the data corresponding to the currently displayed hydrograph with the 'download data' links
+                nodeElem.select('#iv-download-container').call(link(store, (container, {currentIVDateRange, parameterCode, showIVTimeSeries, queryInformation}) => {
```
mbucknell, line 172: I would like this to follow the same pattern that we have already established. Put the rendering code in a separate module and follow the pattern already established for other sub-components. https://github.com/usgs/waterdataui/blob/master/assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js is a nice simple example. The code will be simpler, less prone to errors as things change, and it can be more thoroughly tested. Let's talk if you have questions or comments.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -171,6 +171,7 @@ export const plotSeriesSelectTable = function(elem,
             tr.append('td')
                 .append('a')
                     .attr('href', param => `${config.WATERALERT_SUBSCRIPTION}/?site_no=${siteno}&parm=${param.parameterCode}`)
+                    .attr('class', 'usa-tooltip usa-link')
```
mbucknell, line 174: This should not have usa-tooltip as a style. It's not a tooltip.

assets/src/styles/pages/_monitoring-location.scss
```diff
@@ -18,7 +18,7 @@
   @include u-padding-bottom(1);
 }
 
-#dv-accordion {
+.usa-accordion {
```
mbucknell, line 21: Changing this will cause all accordions to have that margin and that may not be desirable.

wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -9,6 +9,38 @@
...
         </div>
         <div class="select-time-series-container"></div>
+        <div class="graph-data usa-accordion">
+            <h2 class="usa-accordion__heading">
+                <button class="usa-accordion__button" aria-expanded="false" aria-controls="iv-download-container">
+                    Download this graph's data
+                </button>
+            </h2>
+            <div id="iv-download-container" class="usa-accordion__content usa-prose">
+                <ul class="usa-fieldset usa-list--unstyled">
```
mbucknell, line 19: This is the part I would render in code rather than here.

mbucknell (changes requested):

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: Confused about this one. I thought I was following the guidance here: https://designsystem.digital.gov/components/tooltip/ .

assets/src/styles/pages/_monitoring-location.scss
abriggs-usgs: Is there a reason we would like the accordions to be different? There were several different margins applied, some using 'em' and some using USWDS units, making them slightly uneven in spacing. This standardizes them and can always be overridden if the odd case arises.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
mbucknell: Is it supposed to have a tooltip? If it is, you are rendering dynamically and will need to initialize the tooltip (see how it is done in d3render/tooltips.js. I'm going to update that one because you really should pass the element to which you are adding the tooltip. So here it would be components.tooltip.init(tr.node()).

wdfn-server/waterdata/templates/macros/components.html
abriggs-usgs: Do you mean the entire list? I did it this way for a couple reasons, the main one was that this keeps the list order intact. Even though the 'compare' and 'median' only exist when the user selects them, I wanted them to always appear in the same order no matter if the user selected median first and compare second. That would be difficult to control in JavaScript. The second reason, was I thought you mentioned in a previous review that we should do as much structure in the template and as little creation in JavaScript as possible.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: Oooh, that would be why the tooltip didn't have the behavior I was expecting -- I thought it was just a browser display difference. And yes, the product owners, added that they wanted a tooltip added, after you did your first review.

assets/src/styles/pages/_monitoring-location.scss
mbucknell: This particular one has been put is specifically to style the dv-accordion. (Groundwater Data). I would be fine with removing it entirely if it does not affect the display of the dvHydrograph.

assets/src/scripts/monitoring-location/components/hydrograph/index.js
abriggs-usgs: Makes sense, I'll fix it up.

mbucknell (approved):
Nice tests for download-links.

PR #1265, opened 2020-11-18
Relative imports

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run
- [ ] Update the changelog appropriately (The changelog doesn't need to be updated for this internal code change)

Title
-----------
Use relative imports when importing a module within the same directory. The named aliases are used when importing a module from a different directory. This change affects the .spec.js files the most. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/d3-rendering/tooltips.js
+++ b/assets/src/scripts/d3-rendering/tooltips.js
@@ -7,7 +7,7 @@
  */
 
 // Required to initialize USWDS components after page load
-import components from '../../node_modules/uswds/src/js/components';
+import components from 'uswds/src/js/components';
 
 export const appendTooltip = function(elem, text) {
 	let tooltip = elem.append('div')
--- a/assets/src/scripts/d3-rendering/tooltips.spec.js
+++ b/assets/src/scripts/d3-rendering/tooltips.spec.js
@@ -1,8 +1,9 @@
 import {select} from 'd3-selection';
-import {appendTooltip} from 'ui/tooltips';
+
+import {appendTooltip} from './tooltips';
 
 
-describe('tooltips', () => {
+describe('d3-rendering/tooltips', () => {
 
     let testDiv;
     beforeEach(() => {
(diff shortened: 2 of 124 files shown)


assets/src/scripts/d3-rendering/tooltips.js
```diff
@@ -7,7 +7,7 @@
  */
 
 // Required to initialize USWDS components after page load
-import components from '../../node_modules/uswds/src/js/components';
+import components from 'uswds/src/js/components';
```
abriggs-usgs, line 10: Is this an 'alias-plugin' thing, or just a standard Rollup thing?

assets/src/scripts/d3-rendering/tooltips.spec.js
```diff
@@ -1,8 +1,9 @@
 import {select} from 'd3-selection';
-import {appendTooltip} from 'ui/tooltips';
 
+import {appendTooltip} from './tooltips';
 
-describe('tooltips', () => {
+
+describe('d3-rendering/tooltips', () => {
```
abriggs-usgs, line 6: Is just to tell us which tooltips are being tested, or is there greater significance?

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/selectors/legend-data.spec.js
```diff
@@ -1,6 +1,6 @@
-import {lineMarker, rectangleMarker, textOnlyMarker} from 'd3render//markers';
```
abriggs-usgs, line 1: How did this still work??

abriggs-usgs (approved):
Thanks for cleaning that up. Much improved.

assets/src/scripts/d3-rendering/tooltips.js
mbucknell: Just standard rollup. I'm not sure why I didn't try it this way before.

assets/src/scripts/d3-rendering/tooltips.spec.js
mbucknell: Yeah. I have been trying to change these so it's easier in the test output to figure out where the error occurred. In this particular case we have tooltips in d3-rendering and the then tooltip in hydrograph and dv-hydrograph.

assets/src/scripts/monitoring-location/components/daily-value-hydrograph/selectors/legend-data.spec.js
mbucknell: You mean with the double slashes? I think it just ignored it. Some things seem to be picky about that put clearly the rollup code is not .

PR #1276, opened 2020-11-25
Wdfn 445 p2 addon additional links

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Add Additional Links to Download Links section
-----------
Adds a couple of pertinent links to the download links area

After making a pull request
---------------------------

### DIFF
--- a/assets/src/styles/pages/_monitoring-location.scss
+++ b/assets/src/styles/pages/_monitoring-location.scss
@@ -18,10 +18,6 @@
   @include u-padding-bottom(1);
 }
 
-.iv-graph-data-download-container {
-  @include u-padding(1);
-}
-
 table {
   #site-summary {
     table-layout: fixed;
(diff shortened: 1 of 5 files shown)


assets/src/styles/pages/_monitoring-location.scss
```diff
@@ -18,6 +18,10 @@
   @include u-padding-bottom(1);
 }
 
+.iv-graph-data-download-container {
```
mbucknell, line 21: This style is for a class, but line 18 in the macro shows an id. Either this is wrong or you have a class with the same name as an id. I'm not crazy about the latter.

mbucknell (approved):
See the comment about iv-graph-data-download-container

assets/src/styles/pages/_monitoring-location.scss
abriggs-usgs: Wow. Good catch. I think this was a leftover from some changes that were later changed (again). I removed the lines, since they are no longer needed.

PR #1285, opened 2020-12-01
Wdfn 391 - Show other mon loc sites on map, not just upstream/downstream

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Show nearby active monitoring locations on monitoring location page

Description
-----------
The active sites must be in the RTN or RTS networks. When the user changes the map extent, new service calls are made to fetch the active sites within the RTN and RTS networks and the active site layer updated. Each site is clickable, showing an information popup, including a clickable link to that site's page.

I also used this as an opportunity to unify the code which calls the observations endpoint, consolidating into a single javascript web-service module and endpoint. Also changed names from 'data' and 'sites' to 'monitoringLocations' as appropriate to make it more explicit what the variable represents.

I also updated the NLDI endpoints to use the latest API rather than deprecated interfaces. We are no longer retrieving the upstream/downstream sites from NLDI.

After making a pull request
---------------------------


assets/src/scripts/d3-rendering/info-tooltip.js
```diff
@@ -18,5 +18,5 @@ export const appendInfoTooltip = function(elem, text) {
         .attr('class', 'fas fa-info-circle');
 
     // Need to initialize USWDS tooltip explictly after page load
-    components.tooltip.init(elem.node());
+    components.tooltip.on(elem.node());
```
mbucknell, line 21: USWDS just updated their guidance to include a description on how to dynamically initialize their components and recommend that the "on" function is used.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -260,10 +260,11 @@ export const drawDateRangeControls = function(elem, store, siteno) {
             .attr('type', 'text')
             .attr('aria-describedby', 'custom-end-date-label custom-end-date-hint');
 
-        // required to init the USWDS date picker after page load
-        components.datePicker.init(elem.node());
+        // required to init the USWDS date picker after page load before calling the
+        // dateRangePicker on function
+        components.datePicker.init(dateRangePicker.node());
```
mbucknell, line 265: I had to use "init" for the datePicker part of the dateRangePicker initialization. Not sure why. I reported to USWDS and others do not have the same issue and can use "on"

assets/src/scripts/monitoring-location/components/map/legend.js
```diff
@@ -34,30 +26,50 @@ const fetchLayerLegend = function(layer, defaultName) {
         });
 };
 
-export const addMonitoringLocationMarker = function(legendControl) {
-        const siteLegendList = select(legendControl.getLegendListContainer()).append('ul')
+/*
+ * Draws a monitoring location marker within the legendListContainer
+ * @param {D3 selection} legendListContainer
+ */
+export const drawMonitoringLocationMarkerLegend = function(legendListContainer) {
```
mbucknell, line 33: I changed the methods here to take the legendListContainer so that they are not dependent on Leaflet at all. This makes it easier to test the functions and it follows the pattern for most of our D3 rendering code.

assets/src/scripts/monitoring-location/components/map/legend.js
```diff
@@ -75,42 +87,6 @@ export const createFIMLegend = function(legendControl, isFIMAvailable) {
...
-        select(legendControl.getContainer()).select('#fim-legend-list').remove();
-    }
-};
-
-/**
- * Creates the NLDI legend if NLDI data is available, otherwise removes the NLDI legend if it exists.
- * @param {L.Control} legendControl - Leaflet legend control
- * @param {Boolean} is NLDI available
- */
-export const createNldiLegend = function(legendControl, isNldiAvailable) {
```
mbucknell, line 87: I moved this to nldi-mapping as it needed to share styling constants with the code in that module.

assets/src/scripts/monitoring-location/components/map/nldi-mapping.js
```diff
@@ -77,21 +51,38 @@ export const addNldiLayers = function (map, upstreamFlows, downstreamFlows, upst
       });
     };
 
-    const fetchNldiLinesLayer = function(nldiData, style) {
+    const getNldiLinesLayer = function(nldiData, style) {
```
mbucknell, line 54: I changed the names here from "fetch*" to "get" because we aren't actually making a web service call and our convention is to use "fetch" for that. I think this makes it clearer that it's just transforming the data into a layer object.

assets/src/scripts/monitoring-location/components/map/legend.js
abriggs-usgs: Much nicer 👍

wdfn-server/config.py
```diff
@@ -26,7 +26,7 @@
 
 DEBUG = False
 
-SERVER_SERVICE_ROOT = 'https://waterservices.usgs.gov'  # Used for webserver calls to waterservices. Allows us to use a private url.
+SERVER_SERVICE_ROOT = 'https://waterservices.usgs.gov'  # Used for webserver calls to waterservices.
 SERVICE_ROOT = 'https://waterservices.usgs.gov'  # Use for client side calls to waterservices. Most be a public url.
```
abriggs-usgs, line 30: Confused by why SERVER_SERVICE_ROOT and SERVICE_ROOT have the same value

abriggs-usgs (approved):
Wow. Lots of stuff to take in.  To my eye everything looks reasonable. This will be a neat new feature.

wdfn-server/config.py
mbucknell: So the server one is server side as the name implies and the comment states whereas SERVICE_ROOT is for client side. This allows us to use internal only URLs for fetches performed by the server which may have performance benefits. The client side (at least for the public) must be a publicly available URL.

abriggs-usgs: Hmm. So https://waterservices.usgs.gov has both internal and external URLS? I thought the external vs internal was a property set at the domain level.

mbucknell: By default we are using the same URL (makes local development easier). But in production, these are not the same.

PR #1290, opened 2020-12-03
Wdfn 392 help mobile users not get trapped in map

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-392 Help Mobile Users Not Get Trapped in Map
-----------
Prevents users of mobile devices from getting 'stuck' in map when scrolling.

Description
-----------
This uses some basic Leaflet map options to disable one finger scrolling of the map for mobile devices. It still allows users to use pan and zoom with two fingers, which is now a standard implementation on most Google maps. 

After making a pull request
---------------------------


assets/README.md
```diff
@@ -60,3 +60,62 @@ run:
...
+
+Install a 'tunneling' application (LocalTunnel is an example (this is not an endorsement)).
+
+Start the tunneling application; make a tunnel for each server
+
+```bash 
+An example using LocalTunnel 
+(each command is in a seprate terminal, since each will be a running process)
+lt --port 5050 (for the Flask server)
+lt --port 9000 (for the Static Assets server)
```
mbucknell, line 75: I think if you set CONTAINER_RUN to True you shouldn't need to run the static assets server separately. What WhiteNoise does is enables the Flask server to serve out the static. So I think this may be unnecessary.

assets/README.md
```diff
@@ -60,3 +60,62 @@ run:
...
+cp config.py.sample instance/config.py
+```
+If you don't have an 'instance/config' make one and then replace the value of 'STATIC_ROOT' with
+the URL returned for the Static Assets server. 
+
+Once you have changed the STATIC_ROOT, go to 'wdfn-server/config.py' and find the code
+```bash
+if os.getenv('CONTAINER_RUN', False):
+```
+Change 'False'  to 'True' (noting the capitalization).
```
mbucknell, line 93: I would not change the wdfn-server/config.py. This risks committing the config file with this change. Rather set an environment variable, CONTAINER_RUN to True.

assets/README.md
```diff
@@ -60,3 +60,62 @@ run:
...
+
+Start your local Flask and Static Asset server.
+```bash
+Flask server on port 5050 - from the WDFN-server directory run
+env/bin/python run.py
+
+Static Assets on port 9000 - from the assets direcory run
```
mbucknell, line 100: As stated above, I don't think you need to run the asset server.

mbucknell (approved):
Please see the comments about the readme and running Local Tunnel without the asset server and using an environment variable rather than changing the wdfn-server/config.py.

assets/README.md
abriggs-usgs: That would make things simpler. Testing now . . .

PR #1306, opened 2020-12-11
Wdfn 472 - NWISweb site page links into navigation

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


wdfn-server/waterdata/templates/base.html
```diff
@@ -9,10 +9,14 @@
 {% endblock %}
 
 {% block body %}
-    {% include 'partials/base.html' %}
```
mbucknell, line 12: I removed partials/base.html because it didn't really seem to serve a purpose and having an additional "base.html" template seems confusing.

wdfn-server/waterdata/templates/partials/monitoring_location_header.html
```diff
@@ -0,0 +1,93 @@
+{% extends 'partials/header.html' %}
```
mbucknell, line 1: A somewhat subtle thing with Jinja templating is that if you "include" a template you can not then override a block in the template where you included the partial. So in order to add page_specific_menus, I had to extend (inherit) the partials/header.html and override page_specific_menus in that. 

This means some repetition of template code in monitoring_location.html but it is minimal and consists mostly of including other partials. There was already a lot of overriding in that template.

wdfn-server/waterdata/templates/partials/monitoring_location_header.html
```diff
@@ -0,0 +1,93 @@
...
+        <li class="usa-nav__primary-item">
+            <button class="usa-accordion__button usa-nav__link"
+                    aria-expanded="false"
+                    aria-controls="inventory-section">
+                <span>Data Inventory</span>
+            </button>
+            <ul id="inventory-section" class="usa-nav__submenu">
+                <li class="usa-nav__submenu-item">
+                    <a href="https://waterdata.usgs.gov/nwis/inventory/?site_no={{ stations[0].site_no }}&agency_cd={{ stations[0].agency_cd }}"
+                        target="_blank">
```
abriggs-usgs, line 14: Did we ant to add the 'rel="noopener"' to these links? I don't think there is a big risk here since they all go to pretty trusted URLs, but we have used it in similar situations elsewhere

abriggs-usgs (approved):
Looks good. I think this will be a nice and tidy way of adding these links. And thanks for the added comments for explanation--very helpful!

PR #1313, opened 2020-12-16
Wdfn 336 compare line is behind

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN 336 Compare Line is Behind on One Year Graphs
-----------

Changes to use 'time math' instead of 'calendar math' for calculation of the year to compare so that leap years mesh with normal years. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/web-services/models.js
+++ b/assets/src/scripts/web-services/models.js
@@ -57,9 +57,7 @@ export const getTimeSeries = function({sites, params=null, startDate=null, endDa
         serviceRoot = tsServiceRoot(startDate);
     }
     let paramCds = params !== null ? `&parameterCd=${params.join(',')}` : '';
-
     let url = `${serviceRoot}/iv/?sites=${sites.join(',')}${paramCds}&${timeParams}&siteStatus=all&format=json`;
-
 
     return get(url)
         .then(response => JSON.parse(response))
--- a/assets/src/scripts/web-services/models.js
+++ b/assets/src/scripts/web-services/models.js
@@ -68,9 +68,10 @@ export const getTimeSeries = function({sites, params=null, startDate=null, endDa
 };
 
 export const getPreviousYearTimeSeries = function({site, startTime, endTime, parameterCode}) {
+    const hoursInOneYear = 8760;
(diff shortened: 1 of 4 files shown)


assets/src/scripts/web-services/models.js
```diff
@@ -66,13 +68,11 @@ export const getTimeSeries = function({sites, params=null, startDate=null, endDa
 };
 
 export const getPreviousYearTimeSeries = function({site, startTime, endTime, parameterCode}) {
+    const hoursInOneYear = 8760;
     parameterCode = parameterCode ? [parameterCode] : null;
+    const lastYearStartTime = new DateTime.fromMillis(startTime).minus({hours: hoursInOneYear});
```
mbucknell, line 73: Something I've discovered while doing the jest test conversion: we shouldn't be using new DateTime.fromMillis. The DateTime.fromMillis is a static method on the DateTime class therefore doesn't need the new and in fact this causes an error when running jest tests. So please remove the "new" here and in line 74

assets/src/scripts/web-services/models.spec.js
```diff
@@ -107,16 +109,15 @@ describe('Models module', () => {
 
     describe('getPreviousYearTimeSeries', () => {
         const siteID = '05413500';
-
-        const startDate = new Date('2018-01-02T15:00:00.000-06:00');
-        const endDate = new Date('2018-01-02T16:45:00.000-06:00');
+        const startDate = 1514872800000; // milliSecond version of 01/02/2018
```
mbucknell, line 112: Can we add a test specifically for leap year? This test would have passed before the change.

mbucknell (approved):
See comments. I would like to see a test that specifically covers a leap year.  Please add that but you don't need to send it back to me for review.

assets/src/scripts/web-services/models.js
abriggs-usgs: Ah, very interesting and good to know.

assets/src/scripts/web-services/models.spec.js
abriggs-usgs: I'll do my best

PR #1325, opened 2020-12-22
Wdfn 485

mbucknell:
Please note that the sprite icons in the footer will not appear when using a separate static server as is done typically in development.

### DIFF
--- a/wdfn-server/waterdata/templates/partials/footer.html
+++ b/wdfn-server/waterdata/templates/partials/footer.html
@@ -6,6 +6,40 @@
             <a class="usgs-footer-link" href="https://www2.usgs.gov/laws/accessibility.html">Accessibility</a> |
             <a class="usgs-footer-link" href="https://www.usgs.gov/sitemap">Site Map</a> |
             <a class="usgs-footer-link" href="https://answers.usgs.gov/">Contact USGS</a>
+
+            <div class="social-media-links">
+                <span>Follow</span>
+                <a href="https://twitter.com/usgs" target="_blank">
+                    <svg class="usa-icon usa-icon--size-3" aria-label="USGS twitter" role="img">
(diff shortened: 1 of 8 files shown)


wdfn-server/waterdata/templates/partials/footer.html
```diff
@@ -6,6 +6,40 @@
             <a class="usgs-footer-link" href="https://www2.usgs.gov/laws/accessibility.html">Accessibility</a> |
             <a class="usgs-footer-link" href="https://www.usgs.gov/sitemap">Site Map</a> |
             <a class="usgs-footer-link" href="https://answers.usgs.gov/">Contact USGS</a>
+
+            <div class="social-media-links">
+                <span>Follow</span>
+                <a href="https://twitter.com/usgs" target="_blank">
+                    <svg class="usa-icon usa-icon--size-3" aria-label="USGS twitter" role="img">
+                        <use xlink:href="{{ 'img/sprite.svg#twitter' | asset_url }}"></use>
```
abriggs-usgs, line 14: Not sure if this is an issue or not, but I bumped into at some point in the past. The 'xlink:href' is listed as depreciated: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href.

abriggs-usgs (approved):
Looks like a lot of work for adding icons ;)

wdfn-server/waterdata/templates/partials/footer.html
mbucknell: I asked this question on the uswds-public Slack channel. If you don't use that, it won't work on Safari, so it's not really deprecated.

abriggs-usgs: Well that is kinda dumb, but good to know.

jkreft-usgs: Safari, the new IE

PR #1347, opened 2021-01-11
Iow 495 - Discrete GW Levels are on the IV Graph

mbucknell:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------
Discrete GW Levels are on the IV Graph

Description
-----------
The ground water levels are loaded outside of the components. I did this in preparation for the following on ticket to this to add ground water levels to the dv hydrograph. Since the data will be needed in two places, the component specific module seems to be the wrong place. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/index.js
+++ b/assets/src/scripts/monitoring-location/index.js
@@ -2,9 +2,12 @@ import 'ui/polyfills';
 
 import wdfnviz from 'wdfn-viz';
 
+import config from 'ui/config';
+
 import {getParamString} from 'ml/url-params';
 
 import {configureStore} from 'ml/store';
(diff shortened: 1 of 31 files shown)


assets/src/scripts/monitoring-location/index.js
```diff
@@ -30,6 +50,8 @@ const load = function() {
         }
     });
     let nodes = document.getElementsByClassName('wdfn-component');
+    loadAllGroundWaterData(nodes, store);
```
mbucknell, line 53: we are loading groundwater level data here because it is needed for both the hydrograph and dv-hydrograph components. It is only loaded if at least one of those components is present in the DOM.

abriggs-usgs (approved):
Looks like great work to me.

PR #1349, opened 2021-01-15
Wdfn 467 - Endpoint for the monitoring camera needs to be updated

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Endpoint for the monitoring camera needs to be updated

Description
-----------
The new camera interface has an unofficial call to get the camera metadata so now the application will make that call once, save the results in the app.config and then use that to create the links to the camera data. I also took this opportunity to use html5 elements for the video and to also detect when the video format is not supported by the browser so that the most recent image can be shown instead of a video. 

After making a pull request
---------------------------


wdfn-server/waterdata/templates/macros/components.html
```diff
@@ -153,6 +153,41 @@ <h3>Cooperators</h3>
...
+                        ga-on="click" ga-event-category="accordion" ga-event-action="interactionWithCameraAccordion">
+                    Monitoring Camera
+                </button>
+            </h2>
+            <div id="camera-container" class="usa-accordion__content">
+                {% for camera in cameras %}
+                    <h3>{{ camera.name }}</h3>
+                    <p>{{ camera.description }}</p>
+                    <div>
+                        <video hidden class="med-video camera-video" title="Timelapse Video of {{ camera.name }}" controls="" autoplay="">
```
abriggs-usgs, line 172: Just a question: is there an advantage to using 'controls=""' vs 'controls'? I have seen it both ways.

abriggs-usgs (approved):
Looks well done to me.

wdfn-server/waterdata/templates/macros/components.html
mbucknell: No idea ... I have also seen it both ways. I think I cut and pasted this from the StormSummary page. I think I'll remove the quotes because it seems unnecessary.

PR #1365, opened 2021-01-20
Handle a collection without a timeSeries component

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run
- [ ] Update the changelog appropriately - no changes needed. Bug fix

### DIFF
--- a/assets/src/scripts/monitoring-location/iv-data-utils.js
+++ b/assets/src/scripts/monitoring-location/iv-data-utils.js
@@ -173,21 +173,23 @@ const getConvertedTimeSeries = function(timeSeries, tsRequestKey, parameterCode)
 * all data that appears in the hydrograph. NOTE: The contents of this parameter are mutated
  */
 export const convertCelsiusCollectionsToFahrenheitAndMerge = function(collection) {
-    Object.entries(collection.timeSeries).forEach(([tsRequestKey, timeSeriesDetails]) => {
-        const variableCode = timeSeriesDetails.variable;
-        // Cross reference the 'variableCode' in the 'timeSeries' with 'variables' in the state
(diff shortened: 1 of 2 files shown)


assets/src/scripts/monitoring-location/iv-data-utils.js
```diff
@@ -173,21 +173,23 @@ const getConvertedTimeSeries = function(timeSeries, tsRequestKey, parameterCode)
...
-        // For any Celsius parameter codes, clone the application state variables, 'variables and timeSeries'
-        // and convert the appropriate properties to Fahrenheit. Then add the cloned objects to the 'collection'
-        if (config.TEMPERATURE_PARAMETERS.celsius.includes(parameterCode)) {
-            if (!checkForMeasuredFahrenheitParameters(parameterCode, collection.variables)) {
-                const convertedTimeSeries =  getConvertedTimeSeries(collection.timeSeries[tsRequestKey], tsRequestKey, parameterCode);
-                const convertedVariable =  getConvertedVariable(collection.variables[variableCode]);
-
-                merge(collection.variables, convertedVariable);
-                merge(collection.timeSeries, convertedTimeSeries);
+    if ('timeSeries' in collection) {
```
abriggs-usgs, line 176: Ah, it never occured to me that a collection without a timeseries would exist.

abriggs-usgs (approved):
Good catch!

assets/src/scripts/monitoring-location/iv-data-utils.js
mbucknell: Actually, the majority of sites are like this. You get an empty timeSeries in the response and when normalized that turns into a collection without a timeSeries key.

PR #1366, opened 2021-01-23
Wdfn 440 cleanup user feedback

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-440 Cleanup User Feedback
-----------
Removed old feedback links added new sticky tab at bottom of page that links to feedback form (form yet to be added).

Description
-----------
On desktop . . . 
image

On Mobile . . . 
image

After making a pull request
---------------------------


wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -306,11 +301,15 @@ <h1>Error: HTTP {{ status_code }} -- {{ reason }}</h1>
             {% endif %}
         </div>
 
-        {% if config.TOUCHPOINT_SCRIPT %}
```
mbucknell, line 309: I'm in favor of removing this entirely. If we choose to use something like Touchpoint for feedback it should be on our feedback page. This needs to be removed from the ansible playbook.

mbucknell (changes requested):
TOUCHPOINT_SCRIPT should be removed from config.py and the ansible playbook updated.

wdfn-server/waterdata/templates/monitoring_location.html
abriggs-usgs: Will do 👍

PR #1378, opened 2021-01-29
Iow 853 - add discrete groundwater levels to little brush graph

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Add discrete groundwater levels to little brush graph

Description
-----------
I also fixed an issue where focus circles could potentially appear of of the graph. We needed to filter the set of points used in the getNearestTime function to those that are in the domain of the main graph which can vary as the brush varies.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/discrete-data.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/discrete-data.js
@@ -11,10 +11,13 @@ const GW_LEVEL_CLASS = 'gw-level-point';
  * @param {D3 scale} xScale
  * @param {D3 scale } yScale
  */
-export const drawGroundwaterLevels = function(svg, {levels, xScale, yScale}) {
+export const drawGroundwaterLevels = function(svg, {levels, xScale, yScale, enableClip}) {
     svg.selectAll('#iv-graph-gw-levels-group').remove();
     const group = svg.append('g')
         .attr('id', 'iv-graph-gw-levels-group');
+    if (enableClip) {
+        group.attr('clip-path', 'url(#graph-clip)');
+    }
 
     levels.forEach((level) => {
         group.append('circle')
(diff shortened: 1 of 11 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/discrete-data.js
```diff
@@ -11,10 +11,13 @@ const GW_LEVEL_CLASS = 'gw-level-point';
  * @param {D3 scale} xScale
  * @param {D3 scale } yScale
  */
-export const drawGroundwaterLevels = function(svg, {levels, xScale, yScale}) {
+export const drawGroundwaterLevels = function(svg, {levels, xScale, yScale, enableClip}) {
```
mbucknell, line 14: I did need to add a clip to drawing groundwater levels. We could remove this if the selector only gave us visible points for the graph but I figured there would be unexpected results so just but in the clip.

assets/src/scripts/monitoring-location/components/hydrograph/selectors/cursor.js
```diff
@@ -7,9 +7,12 @@ import {getCurrentMethodID} from 'ml/selectors/time-series-selector';
 
 import {getVisibleGroundWaterLevelPoints} from './discrete-data';
 import {getCurrentVariablePointsByTsId} from './drawing-data';
-import {getMainXScale, getMainYScale} from './scales';
+import {getMainXScale, getMainYScale, getTimeRange} from './scales';
 import {isVisible} from './time-series-data';
 
+const isInTimeRange = function(dateTime, timeRange) {
```
mbucknell, line 13: Utility function check if a dateTime is within a time range. This is used to filter the list of visible points so that we only have the points visible on the desired graph.

abriggs-usgs: That's an nice and straight forward solution.

abriggs-usgs (approved):
looks good to me!

PR #1399, opened 2021-02-12
Iow 681 reference lists auto update

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

IOW-681 Reference Lists Auto Update
-----------
This is part one of a multi-part task.
Here the local lookup files are removed and replaced with a call to pull files from Amazon Web Services. The files are pull when the application starts and then every 24 hours after. This way new changes to the lookups will never be more the 24 hours out of date. 

Description
-----------
This group of changes deal with the Waterdataui side of things. In this pull request the code is modified so that the lookup files are pulled from a S3 bucket rather than from local files. 

Notes
- The local files formerly used as the lookup source have been retained until we can verify that the changes work- they can then be removed. 
- A sub directory has been added to the data folder called 'lookups'. This folder contains backup copies of the lookups so that they are not constantly pulled for S3 every time the local server restarts. This folder is NOT included in Git. 
- In the deployed application, after the lookup files are retrieved from S3 and a backup copy of the file is stored on the server. 
- The application will pull the lookup files from S3 each time it starts and once every 24 hours.
- The backup files are loaded if the application attempts to update the files from S3 but can't.
- If the application is newly deployed there will not be any backup files. So, if the pull from S3 fails at that point the build and deploy will succeed, but the application will error out. This would be unlikely, and an email is sent to the team immediately if it happens.
- The current location of the S3 files is temporary and we will need request a long-term home for them.
-The S3 bucket that hosts the files is public, so there should not be any permissions issues when the application is deployed. 

- This will need additional tests - but wanted to have the prelimiary code reviewed first, before I spend a lot of time on tests that will need changing. 

After making a pull request
---------------------------


wdfn-server/config.py
```diff
@@ -101,8 +91,18 @@
...
+
+DATA_DIR = os.path.join(os.path.dirname(__file__), 'data')
+LOOKUP_ENDPOINTS = {
+    'HUC_LOOKUP': 'https://labs-test-website.s3-us-west-2.amazonaws.com/test-lookups/huc_lookup.json',
+    'COUNTRY_STATE_COUNTY_LOOKUP': 'https://labs-test-website.s3-us-west-2.amazonaws.com/test-lookups/country_state_county_lookup.json',
+    'NWIS_CODE_LOOKUP': 'https://labs-test-website.s3-us-west-2.amazonaws.com/test-lookups/nwis_code_lookup.json'
+}
+
+# Use a comma separated string, with no spaces
+EMAIL_LIST_DEVELOPERS = 'abriggs@contractor.usgs.gov'
```
mbucknell, line 108: Should probably list two developers eventually. Also wondering if we should make this an environment variable so that it can be set per tier.

wdfn-server/waterdata/__init__.py
```diff
@@ -43,18 +47,100 @@ def _create_log_handler(log_directory=None, log_name=__name__):
...
+        server.send_message(msg)
+        server.quit()
+
+    except Exception as e:
+        app.logger.error('Error when sending email about lookups: ', e)
+
+
+def load_lookup_from_backup_file(lookup_name, is_first_load):
+    """
+    Loads the lookup information from a file
```
mbucknell, line 73: Perhaps you could add that it saves the contents in the app.config[lookup_name]

wdfn-server/waterdata/__init__.py
```diff
@@ -43,18 +47,100 @@ def _create_log_handler(log_directory=None, log_name=__name__):
...
+            load_lookup_from_backup_file(lookup, is_first_load = False)
+        except requests.exceptions.RequestException as e:
+            app.logger.error('Request to get lookup file failed for {} with error: {}  '.format(lookup, e))
+            load_lookup_from_backup_file(lookup, is_first_load = False)
+
+
+# When the application is first started, there will be no backup files saved for the lookups and the lookups
+# will not be in memory. So, let's get them.
+# If the backup lookup files already exist, then we can skip this step.
+# NOTE: The lookup files are saved locally but not committed to Git. This stops the application from pulling the
```
mbucknell, line 131: I am not opposed to doing this, but I think it is rare that the application is restarted without a redeployment. Occassionally this happens if we update the config.  But I think it is fine to leave it in.

mbucknell (approved):
Looks good to me

abriggs-usgs:
Changing the implementation on this, so closing this pull request. Changes will be in WDFN-517

PR #1409, opened 2021-02-18
WDFN-503_Divide By Zero Error

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-503_Divide By Zero Error
-----------
FIxed an error related to JavaScript resolving numerical values of 0 to (but not including) 1 as false, and therefore returned a null value rather than the correct calculated value. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/iv-data-utils.js
+++ b/assets/src/scripts/monitoring-location/iv-data-utils.js
@@ -155,7 +155,7 @@ const getConvertedTimeSeries = function(timeSeries, tsRequestKey, parameterCode)
             return {
                 ...point,
                 qualifiers: [...point.qualifiers],
-                value: point.value ? convertCelsiusToFahrenheit(point.value).toFixed(2) : null
+                value: Number.isNaN(point.value) || point.value === '' || isNaN(point.value) ? null : convertCelsiusToFahrenheit(point.value).toFixed(2)
             };
         }),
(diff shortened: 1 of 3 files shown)


assets/src/scripts/monitoring-location/iv-data-utils.js
```diff
@@ -155,7 +155,8 @@ const getConvertedTimeSeries = function(timeSeries, tsRequestKey, parameterCode)
             return {
                 ...point,
                 qualifiers: [...point.qualifiers],
-                value: point.value ? convertCelsiusToFahrenheit(point.value).toFixed(2) : null
+                value: Number.isNaN(point.value) || point.value === '' || isNaN(point.value) ?
```
mbucknell, line 158: Are both Number.isNaN and isNaN needed?

mbucknell (approved):

assets/src/scripts/monitoring-location/iv-data-utils.js
abriggs-usgs: Well, it looks like Number. IsNaN isn't needed. JavaScript can certainly be squarely with truthful evaluations. Number. IsNaN wouldn't return null for a string in the above code, it would return NaN. Which means it evaluated as true, then ran the convert function which returned NaN, Weird. But I was more surprised that an empty string, would return 32 degrees. I guess in JavaScript land a blank string is a numerical zero. Just some of many reasons JavaScript is not idea for calculations.

PR #1413, opened 2021-02-19
Iow 866 add approvals to groundwater levels table

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

IOW-866 Add Approvals to Groundwater Level Table
-----------
Adds a column indicating the approval status of the data to the DV (groundwater) Data Table.

Description
-----------
Looks like this . . .
image

This is a monitoring location with a provisional value, http://localhost:5050/monitoring-location/375259097252901/#parameterCode=72019 , which looks like . . . 
image

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/utils.js
+++ b/assets/src/scripts/utils.js
@@ -290,3 +290,17 @@ export const getNearestTime = function(data, time) {
     // Return the nearest data point and its index.
     return datum;
 };
+
+/*
+*
+*/
+export const approvalCodeText = function(approvalCode) {
+    const approvalText = {
+        P: 'Provisional',
+        A: 'Approved',
+        E: 'Estimated',
+        default: 'unknown code'
+    };
+
+    return approvalText[approvalCode] || approvalText.default;
+};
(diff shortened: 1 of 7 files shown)


assets/src/scripts/utils.js
```diff
@@ -290,3 +290,20 @@ export const getNearestTime = function(data, time) {
     // Return the nearest data point and its index.
     return datum;
 };
+
+/*
+* When given an approval code, will return the text equivalent
+*  @param {String} approvalCode - Usually a letter such as 'A'
+*   @return {String} - an easy to understand text version of an approval code
+*/
+export const approvalCodeText = function(approvalCode) {
```
mbucknell, line 299: Why put this here? It is only used in components/hydrograph/selectors/discrete-data? At this point, approval encoding is different for each kind of data (i.e. iv, dv, and gw) so I don't think it's generally usable.

mbucknell (approved):
See comment. I would prefer seeing the approvalCodeText to leave in the selector/discrete-data.js module.

assets/src/scripts/utils.js
abriggs-usgs: Gotcha. Yeah, I didn't know if there was cross over between the data types so thought it might be a candidate for reuse. WIll move.

PR #1434, opened 2021-03-04
Wdfn 500 download links refactor

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN- 500 Download Links Work with Refactor Code
-----------
Returns the functionality of the links so that the act just as they did before the refactor.

Description
-----------
This adds some behind the scenes improvements as well: the IV data links now download data that is the exact timeframe as on the graph. Previously, it was possible that time frame could be slightly off if the user waited with the page open until the background services updated with new IV points. 

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -53,19 +21,22 @@ const createUrlForDownloadLinks = function(currentIVDateRange, queryInformation,
...
-        currentIVDateRange,
-        parameterCode,
-        showIVTimeSeries,
-        queryInformation,
-        medianData,
-        anyVisibleGroundwaterLevels,
-        requestTimeRange
+        currentTimeRange,
+        priorYearTimeRange,
+        inputs,
```
mbucknell, line 30: The time parameters part of this object aren't used. I wonder if it would be clearer to pass in the parameterCode, showCompareIV and showMedian directly. I was confused that you were grabbing both the inputs and the time ranges (I understand why you need the time ranges though). I don't feel real strongly about this so I'm fine with this staying as is.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -53,19 +21,22 @@ const createUrlForDownloadLinks = function(currentIVDateRange, queryInformation,
...
+        medianStatisticsData,
```
mbucknell, line 31: I wonder if we should make selectors that return true if there is any data of a particular kind. Seems like I have checked this elsewhere too.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -87,43 +58,53 @@ export const renderDownloadLinks = function(elem, store, siteno) {
...
+                        endTime: endDT,
+                        format: 'rdb'
+                    }),
                     gaEventAction: 'downloadLinkCurrent',
                     tooltipText: 'Monitoring location data as shown on graph'
                 });
         }
 
-        if (hasIVData && showIVTimeSeries.compare) {
+        if (inputs.loadCompare && compareIVData && Object.keys(compareIVData.values) > 0) {
```
mbucknell, line 77: Same suggestion as above.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -87,43 +58,53 @@ export const renderDownloadLinks = function(elem, store, siteno) {
...
                 });
         }
 
-        if (showIVTimeSeries.median && medianData) {
+        if (inputs.loadMedian && medianStatisticsData && Object.keys(medianStatisticsData) > 0) {
             listOfDownloadLinks.append('li')
                 .call(createDataDownloadLink, {
                     displayText: 'Median data',
-                    url: `${config.SERVICE_ROOT}/stat/?format=rdb&sites=${siteno}&statReportType=daily&statTypeCd=median&parameterCd=${parameterCode}`,
+                    url: `${config.SERVICE_ROOT}/stat/?sites=${siteno}&statReportType=daily&statTypeCd=median&parameterCd=${inputs.parameterCode}&format=rdb`,
```
mbucknell, line 97: I think you should follow the same pattern as with the IV data and put a get*Url function in statistics-data and groundwater-levels. We know for sure at some point these URLs will change when they are reimplemented. Putting the code in one place will help keep everything easy to change.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.test.js
```diff
@@ -39,33 +39,28 @@ xdescribe('monitoring-location/components/hydrograph/download-links', () => {
...
-                                'value': '00060'
-                            }
-                        },
-                        '450807142': {
-                            'variableCode': {
-                                'value': '00010'
-                            }
-                        }
-                    }
+                'hydrographData': {
```
mbucknell, line 42: You might consider creating a single TEST_STATE using some of the constants in mock-hydrograph-state and then use the spread operator to customize for each test. This is the pattern I used for most of the tests in components/hydrograph.

assets/src/scripts/monitoring-location/components/hydrograph/index.js
```diff
@@ -20,8 +20,8 @@ import {Actions as floodDataActions} from 'ml/store/flood-inundation';
 
 import {showDataLoadingIndicator} from './data-loading-indicator';
 import {drawDataTables} from './data-table';
+import {renderDownloadLinks} from './download-links';
```
mbucknell, line 23: Can we change the name to drawDownloadLinks to match the other functions?

mbucknell (changes requested):
I have a few suggestions. I would like to see get*URL calls added to statistics-data and groundwater-levels in web-services so that everything follows the same pattern and changes to the API will only need to be made in one place.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
mbucknell: Actually thinking about this, you could use the time portion of inputs directly to get the url for primary. You would still need to get the priorYearTimeRange to get the right time range for the compare IV data though. But these are only suggestions. There are obviously pros and cons both ways.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
```diff
@@ -87,43 +58,53 @@ export const renderDownloadLinks = function(elem, store, siteno) {
...
+                        endTime: DateTime.fromMillis(priorYearTimeRange.end).toISO(),
+                        format: 'rdb'
+                    }),
                     gaEventAction: 'downloadLinkCompare',
                     tooltipText: 'Data from last year with the same duration as in graph'
                 });
         }
 
-        if (showIVTimeSeries.median && medianData) {
+        if (inputs.loadMedian && medianStatisticsData && Object.keys(medianStatisticsData) > 0) {
```
mbucknell, line 93: Same suggestion as above.

assets/src/scripts/monitoring-location/components/hydrograph/index.js
abriggs-usgs: Sure can!

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
abriggs-usgs: lol. I knew there was another way to do this, but was blanking on it.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
abriggs-usgs: I debated about this. To me, it seemed like an extra computing step, since in both cases the function has to check the state. But I'm okay with the idea, if you think it works better.

mbucknell: I don't feel really strongly about it. Maybe once this code is merged in, I'll do a check to see if I think it's worth it. I think this is fine for now.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
abriggs-usgs: I hesitated on moving these because they only get used once, but following the pattern makes sense. Will move.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.test.js
abriggs-usgs: I used a test state for some and cleaned out some unneeded items of for others. Hopefully, it as lean as it can be.

assets/src/scripts/monitoring-location/components/hydrograph/download-links.js
abriggs-usgs: I went back and forth on this one. I actually changed the code and then changed it again, then slept on it to see if I felt the same way in the morning :). In the end, I opted to keep the currentTimeRange and the priorYearTimeRange as that seemed a thematically consistent pattern for getting the time frames. I am wondering if we should add start and end time for the compare to getInputsForRetrieval, or maybe split the 'time' selectors off into their own. I like the one-stop-shopping convenience of the getInputsForRetrieval, but don't want to overload its function.

mbucknell: I'd have to think about that. getInputsForRetrieval was meant to essentially be a short cut for getting the inputs for retrieveHydrographData which gets called in multiple places so adding to it doesn't seem right to me. I'm fine with leaving it as you wrote it.

assets/src/scripts/web-services/groundwater-levels.js
```diff
@@ -2,6 +2,19 @@
...
+
+/**
+ *  Formats a URL for the purpose of downloading groundwater levels data in RDB format
+ * @param  {String}  siteno - A code that uniquely identifies a monitoring location
+ * @param  {String}  parameterCode - USGS five digit parameter code
+ * @param {String} startTimeDT - ISO 8601 time
+ * @param {String} endDT- ISO 8601 time
+ * @return {String} The URL used to contact waterservices.usgs.gov
+ */
+export const getServiceURLSGroundwater = function({siteno, parameterCode, startDT, endDT}) {
```
mbucknell, line 14: I think you should use this in the fetch function too. Which means passing in period but that can be defaulted to null. This makes everything more DRY

assets/src/scripts/web-services/statistics-data.js
```diff
@@ -6,6 +6,16 @@ import {parseRDB} from 'ui/utils';
 
 const SERVICE_ROOT = config.SERVICE_ROOT || 'https://waterservices.usgs.gov/nwis';
 
+/**
+ *  Formats a URL for the purpose of downloading median statistics in RDB format
+ *  @param  {String}  siteno - A code that uniquely identifies a monitoring location
+ *  @param  {String}  parameterCode - USGS five digit parameter code
+ *  @return {String} The URL used to contact waterservices.usgs.gov
+ */
+export const getServiceURLStatistics = function({siteno, parameterCode}) {
```
mbucknell, line 15: As with the groundwater levels, use this function in the fetch function to form the URL.

mbucknell (approved):
I'd like to see the get*URL functions used in the fetch* functions to form the URL. This is what will make the code more DRY. You don't need to put this through review again.

assets/src/scripts/web-services/statistics-data.js
abriggs-usgs: Oh, duh. Will do.

PR #1440, opened 2021-03-08
Wdfn 508 - Default methods to the one with data

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Default methods to the one with data

Description
-----------
We are determining the preferred method by picking the method that has the most points. If more than one has the most points, then the one chosen is the one with the most recent data point.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
@@ -3,17 +3,43 @@
  */
 
 import {select} from 'd3-selection';
-import {createStructuredSelector} from 'reselect';
 
 import{link}  from 'ui/lib/d3-redux';
 
 import {getPrimaryMethods} from 'ml/selectors/hydrograph-data-selector';
 import {getSelectedIVMethodID} from 'ml/selectors/hydrograph-state-selector';
+
 import {setSelectedIVMethodID} from 'ml/store/hydrograph-state';
 
-export const drawMethodPicker = function(elem, store, initialTimeSeriesId) {
+import {getPreferredIVMethodID} from './selectors/time-series-data';
+
+const updateAvailableMethods = function(selectElem, methods, store) {
+    selectElem.selectAll('option').remove();
+    if (!methods || !methods.length) {
+        return;
+    }
+    let selectedMethodID = getSelectedIVMethodID(store.getState());
+    const availableMethodIDs = methods.map(data => data.methodID);
+    if (selectedMethodID && !availableMethodIDs.includes(selectedMethodID)) {
+        selectedMethodID = getPreferredIVMethodID(store.getState());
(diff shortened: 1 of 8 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
```diff
@@ -25,32 +55,8 @@ export const drawMethodPicker = function(elem, store, initialTimeSeriesId) {
     pickerContainer.append('select')
         .attr('class', 'usa-select')
         .attr('id', 'method-picker')
+        .call(link(store, updateAvailableMethods, getPrimaryMethods, store))
```
mbucknell, line 58: I pulled this code into a named function. I think we use anonymous functions too often in our rendering code (I have been as guilty of this as anyone else). The anonymous function on line 46 is fine because it's a one line function and what it does is pretty clear. But this one is more complicated and I think worthy of being a named function.

assets/src/scripts/monitoring-location/components/hydrograph/selectors/time-series-data.js
```diff
@@ -77,3 +77,33 @@ export const getPrimaryParameterUnitCode = createSelector(
     getPrimaryParameter,
     parameter => parameter ? parameter.unit : null
 );
+
+/*
+ * Returns a Redux select which returns the IV data method id with the most points. If more than
+ * one method has the same point count, then the time series with the most recent point is chosen.
+ * @returns {Function} which returns {String} the preferred method ID.
+ */
+export const getPreferredIVMethodID = createSelector(
```
mbucknell, line 86: I decided to put this bit of logic in a selector rather than in the method-picker module. It allows us to more completely and easily test the logic.

assets/src/scripts/monitoring-location/selectors/hydrograph-data-selector.js
```diff
@@ -44,12 +44,21 @@ export const getIVValueRange = memoize(dataKind => createSelector(
             return null;
         }
 
-        let values = [];
-        Object.values(ivData.values).forEach(byMethodID => {
-            values.push(...byMethodID.points.filter(point => point.value !== null).map(point => point.value));
-        });
-        if (values.length) {
-            return [Math.min(...values), Math.max(...values)];
+        const result = Object.values(ivData.values).reduce((minMaxAccumulator, byMethodID) => {
```
mbucknell, line 47: I had to change the implementation because the call to Math.min and Math.max fails with a Max Stack exceeded error in the error if the number of values is large (specifically for multi year time ranges).

abriggs-usgs (approved):
Looks good to me and seems to work well. I had to pull down the code to really understand how the 'sort' functions were working.  Good idea to sort by points and time. 👍

PR #1442, opened 2021-03-11
Wdfn 514 rework questions comments button

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-514 Rework Questions and Comments Button
-----------
User testing indicated that the floating feedback, when located to the far right, was utterly invisible to users. So this tries something different. 

Description
-----------
To help users see the feedback button, it is now moved to a centered location. Additionally, a second link has been added to the link section of the header so that we can user test and see which location is the most visible to users. 

Image of links
image

After making a pull request
---------------------------

### DIFF
--- a/wdfn-server/waterdata/templates/monitoring_location.html
+++ b/wdfn-server/waterdata/templates/monitoring_location.html
@@ -305,7 +305,7 @@
     <div class="user-feedback">
         <a href="{{ url_for('questions_comments', email_for_data_questions=email_for_data_questions) }}"
            target="_blank"
-           ga-on="click" ga-event-category="navigation" ga-event-action="toFeedBackForm">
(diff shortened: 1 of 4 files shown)


wdfn-server/waterdata/templates/monitoring_location.html
```diff
@@ -305,7 +305,7 @@ <h1>Error: HTTP {{ status_code }} -- {{ reason }}</h1>
     <div class="user-feedback">
```
mbucknell, line 305: This thought just occurred to me. I wonder if this should go on all pages. The networks pages are public so wondering if this bit of html should be part of the footer? Let's chat with Shawna about this at stand-up. It will be good to distinguish which button they are clicking.

mbucknell (approved):
There are also conflicts that need to be resolved with the Changelog.

PR #1446, opened 2021-03-12
Wdfn 512 change order of custom date

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-512 Change Order of Custom Date
-----------
This switched the display order of the 'days before today' and the 'calendar days' options in the 'custom' time range selection.

Description
-----------
The main issue I ran into here was that the 'days before today' option is of the same type as all of the other standard time range selections ('P7D', 'P30D', 'P1Y) except 'custom'. This was only a issue when the user first selected 'custom' because the menu defaulted to 'days before to day' using 'not custom' as the default because the only way that would not be true is if the user manually entered a set of dates in the URL which would cause the menu to automatically open to show dates menu with the chosen dates added to the menu. So fixing this required adding a way to allow the application to default to the date selection menu when the 'custom' radio button is clicked.

Additionally there are a couple of small clean-up items added:
1) This covers a basic version of WDFN-513, which just changes the name of the 'custom' radio button and adds a bit of CSS to allow the USWDS label to accept more characters without text wrapping.
2) This adds the floating feedback link to the 'footer' template so that it will now appear on all the pages using the 'footer' template. 
3) Also removes the old email addresses that are no longer used from the config.py

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -23,44 +23,51 @@ const DATE_RANGE = [{
...
-    },
-    {
-        id: 'custom-input-calendar-days',
-        value: 'calendar',
-        text: 'Calendar days'
     }
 ];
 
+// The standard 'selectedDateRange' options -- the options that are not a custom period, such as P129D or P3D
+const STANDARD_INITIAL_VALUES = ['custom', 'P7D', 'P30D', 'P1Y'];
```
mbucknell, line 43: We don't use P1Y for 1 year, it is P365D.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -23,44 +23,51 @@ const DATE_RANGE = [{
...
     return isCustomPeriod(dateRange) || isCustomDateRange(dateRange);
-};
+}
 
 /*
  * Set custom form hidden attribute and clear custom container inputs if it is being hidden
  * @param {Boolean} showCustom
  */
-const setCustomFormVisibility = function(showCustom) {
+function setCustomFormVisibility(showCustom) {
```
mbucknell, line 63: I was inconsistant with my helper functions. Let's always do const funcName = function(....)

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -161,14 +171,14 @@ const drawCustomRadioButtons = function(container, initialDateRange) {
...
- * @param {String} initialDateRange - Used to set the initial contents of the text field.
+ * @param {String} initialDateRange - Either in the form of P<some number>P(or Y). Used to set the initial contents of the text field.
  */
 const drawCustomDaysBeforeForm = function(container, store, siteno, initialDateRange) {
     const formContainer = container.append('div')
         .attr('id', 'ts-custom-days-before-today-select-container')
         .attr('class', 'usa-form')
         .attr('aria-label', 'Custom date by days before today specification')
-        .attr('hidden', initialDateRange === 'custom' ? true : null);
+        .attr('hidden', STANDARD_INITIAL_VALUES.includes(initialDateRange) ? true : null);
```
mbucknell, line 181: isn't this the same as !isCustomPeriod(initialDateRange)? If it is, I don't think we need STANDARD_INITIAL_VALUES.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -23,44 +23,51 @@ const DATE_RANGE = [{
...
 const isCustomPeriod = function(dateRange) {
     return dateRange !== 'custom' && !DATE_RANGE.find(range => range.period === dateRange);
 };
 const isCustomDateRange = function(dateRange) {
     return dateRange === 'custom';
 };
-const showCustomContainer = function(dateRange) {
+
+/*
+* Helper function that
```
mbucknell, line 53: Might want to finish the sentence.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -137,9 +144,11 @@ const drawCustomRadioButtons = function(container, initialDateRange) {
         .attr('id', d => `${d.value}-input`)
         .attr('class', 'usa-radio__input')
         .property('value', d => d.value)
-        .property('checked', d => d.value === 'calendar' ? initialDateRange === 'custom' : initialDateRange !== 'custom')
+        .property('checked', d => d.value === 'calendar' ?
+            STANDARD_INITIAL_VALUES.includes(initialDateRange) :  !STANDARD_INITIAL_VALUES.includes(initialDateRange))
```
mbucknell, line 148: See comment below about isCustomPeriod

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -137,9 +144,11 @@ const drawCustomRadioButtons = function(container, initialDateRange) {
...
         .attr('ga-on', 'click')
-        .attr('aria-expanded', d => d.value === 'calendar' ? initialDateRange === 'custom' : initialDateRange !== 'custom')
+        .attr('aria-expanded', d => d.value === 'calendar' ?
+            STANDARD_INITIAL_VALUES.includes(initialDateRange) : !STANDARD_INITIAL_VALUES.includes(initialDateRange))
```
mbucknell, line 151: Ditto

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
```diff
@@ -286,10 +296,10 @@ const drawCustomCalendarDaysForm = function(container, store, siteno, initialDat
         .attr('role', 'customdate')
         .attr('class', 'usa-form')
         .attr('aria-label', 'Custom date specification')
-        .attr('hidden', initialDateRange === 'custom' ? null : true);
-
+        .attr('hidden', STANDARD_INITIAL_VALUES.includes(initialDateRange) ? null : true);
```
mbucknell, line 299: See above comment about isCustomPeriod

mbucknell (approved):
There are some cleanup things that need to be done in date-controls and look at the function isCustomPeriod rather than defining STANDARD_INITIAL_VALUES

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Yup. Thanks for pointing that out!

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Fixed

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Why yes it is. Good point! Using it does result in a number of double negative Booleans, which I always find confusing, but it does make sense to use isCustomPeriod for consistency.

assets/src/scripts/monitoring-location/components/hydrograph/date-controls.js
abriggs-usgs: Ugh, yes, I was going to write this documentation, and then I realized I didn't have the right words to describe what was happening in this really simple function, so I paused and subsequently forgot about it. I added a comment, to it now, but it seems like a lot of words for a simple thing, which may mean there is a better way to organize how these selection menus work. And since, we while probably change how is part of the application works in the near future it may be a good thing to think about then.

PR #1456, opened 2021-03-16
fix feedback links for networks pages

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-512 Change Order of Custom Date (Fix links to feedback form on Network pages) 
-----------
I realized that there was a problem with the way the links to the Feedback form worked on the Network pages. The gist of it is, the feedback form that is called from a Monitoring Location page needs an email address that is regionally specific to that monitoring location. Since there is no one monitoring location on a Networks page, there is no region specific email associated with that page. 

A second issue is that the radio button selections have wording tailored to the giving feedback on for monitoring location page which didn't make much sense for feedback on a Networks page. 

Description
-----------
To deal with the above issues
1) I added a query parameter called 'page_type' that can tell the feedback form which page requested it.
2) On Network pages the URL for the feedback page uses the Default NWIS support email.
3) On the Feedback form called from a Networks page the option to ask a question about the data  has been removed.
4) Conditional Logic was added to the creation of the feedback link in both the header and the footer so that those links will only show on Monitoring Location and Networks pages. 

After making a pull request
---------------------------

### DIFF
--- a/wdfn-server/waterdata/templates/partials/footer.html
+++ b/wdfn-server/waterdata/templates/partials/footer.html
@@ -1,6 +1,6 @@
 {% if request.endpoint == 'monitoring_location' or request.endpoint == 'networks'%}
     <div class="user-feedback">
-        <a href="{{ url_for('questions_comments', page_type=page_type, email_for_data_questions=email_for_data_questions) }}"
+        <a href="{{ url_for('questions_comments', referring_page_type=referring_page_type, email_for_data_questions=email_for_data_questions) }}"
            target="_blank"
            ga-on="click" ga-event-category="navigation" ga-event-action="toFeedBackForm_floatingLink">
             <p class="usa-prose">Questions or Comments</p>
--- a/wdfn-server/waterdata/templates/partials/footer.html
+++ b/wdfn-server/waterdata/templates/partials/footer.html
@@ -1,10 +1,12 @@
-<div class="user-feedback">
-    <a href="{{ url_for('questions_comments', email_for_data_questions=email_for_data_questions) }}"
-       target="_blank"
-       ga-on="click" ga-event-category="navigation" ga-event-action="toFeedBackForm_floatingLink">
-        <p class="usa-prose">Questions or Comments</p>
-    </a>
-</div>
+{% if request.endpoint == 'monitoring_location' or request.endpoint == 'networks'%}
+    <div class="user-feedback">
+        <a href="{{ url_for('questions_comments', page_type=page_type, email_for_data_questions=email_for_data_questions) }}"
+           target="_blank"
+           ga-on="click" ga-event-category="navigation" ga-event-action="toFeedBackForm_floatingLink">
+            <p class="usa-prose">Questions or Comments</p>
+        </a>
+    </div>
+{%  endif %}
 <footer class="usa-footer usa-footer-slim" role="contentinfo">
     <nav class="usa-footer-primary-section content-container" aria-label="Footer Navigation">
         <div class="usgs-footer-nav">
--- a/wdfn-server/waterdata/templates/partials/footer.html
+++ b/wdfn-server/waterdata/templates/partials/footer.html
@@ -1,4 +1,4 @@
-{% if request.endpoint == 'monitoring_location' or request.endpoint == 'networks'%}
+{% if has_feedback_link(request.endpoint) %}
     <div class="user-feedback">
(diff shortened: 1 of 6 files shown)


wdfn-server/waterdata/templates/partials/footer.html
```diff
@@ -1,10 +1,12 @@
-<div class="user-feedback">
-    <a href="{{ url_for('questions_comments', email_for_data_questions=email_for_data_questions) }}"
-       target="_blank"
-       ga-on="click" ga-event-category="navigation" ga-event-action="toFeedBackForm_floatingLink">
-        <p class="usa-prose">Questions or Comments</p>
-    </a>
-</div>
+{% if request.endpoint == 'monitoring_location' or request.endpoint == 'networks'%}
```
mbucknell, line 1: This is fine for now. I think if we begin to add other pages, we may want to implement a context processor on the flask side to insert a variable that enables/disables the feedback section. But this works so let's leave it.

wdfn-server/waterdata/templates/questions_comments.html
```diff
@@ -30,13 +30,18 @@ <h3 id="questions-comments-subtitle">Questions or Comments</h3>
...
+                    <div class="usa-radio">
+                        <input class="usa-radio__input" id="feedback-selection-contact" type="radio" name="feedback-type" value="contact" required>
+                        <label class="usa-radio__label" for="feedback-selection-contact">Contact someone about data at this location</label>
+                    </div>
+                {% endif %}
                 <div class="usa-radio">
                     <input class="usa-radio__input" id="feedback-selection-report" type="radio" name="feedback-type" value="report">
-                    <label class="usa-radio__label" for="feedback-selection-report">Report a problem with this location</label>
+                    <label class="usa-radio__label" for="feedback-selection-report">
+                        Report a problem{%  if page_type ==  "monitoring" %} with this location {% endif %}
```
mbucknell, line 43: Can we call the variable referring_page_type? I got a little confused here with the changes you made above that used request.endpoint. I think just changing the variable name helps.

mbucknell (approved):
One suggestion but looks good to go!

wdfn-server/waterdata/templates/questions_comments.html
abriggs-usgs: Sure can.

wdfn-server/waterdata/templates/partials/footer.html
abriggs-usgs: Sounds good. Reading about Context Processors right now . . .

mbucknell: it's just a way to pass a variable to any page. We used a context processor to pass whether a user was authenticated: https://github.com/usgs/pubswh-ui/blob/5ede5634451681f1584885da49d46d6a77803fd7/server/pubs_ui/auth/views.py#L58

abriggs-usgs: Okay, I think I got the context processor working correctly. A pretty handy feature.

mbucknell: Yes it is.Bonus is that you can write tests for it too. I didn't expect that you would do this but if you added it, it should really have tests

wdfn-server/waterdata/views.py
```diff
@@ -23,6 +23,22 @@
 
 NWIS = NwisWebServices(app.config['SERVER_SERVICE_ROOT'], app.config['SITE_SERVICE_CATALOG_ROOT'])
 
+
+def has_feedback_link(request_endpoint):
```
mbucknell, line 27: I am a little surprised this works. I would think you just call request.get('endpoint') or something like that. How does it get request_endpoint?

wdfn-server/waterdata/templates/partials/footer.html
abriggs-usgs: Will do.

wdfn-server/waterdata/views.py
abriggs-usgs: lol. I knew I should have you review it. I pass 'request.endpoint' in, in the 'if' in the template.

mbucknell: I'm okay with not using a context processor too. It's just something to consider for the future. Maybe just revert?

PR #1463, opened 2021-03-23
Wdfn 524 - Rework download data section

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Rework download data section

Description
-----------
Added a section ("select actions") above the parameter selection list, that contains the change time span and download data sections of the page. The select actions will show the selected section when clicking on the appropriate button. Only one section can be viewed at a time. The time span section was moved as is and will be updated in a subsequent JIRA. The download section was reworked to have a set of checkboxes that a user will check to select the data that they are interested in. Clicking the download button opens a new tab for each selected kind of data.

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/download-data.test.js
```diff
@@ -0,0 +1,131 @@
...
+                window.requestAnimationFrame(() => {
+                    div.select('input[value="primary"]').property('checked', true);
+                    div.select('input[value="groundwater-levels"]').property('checked', true);
+                    div.select('input[value="median"]').property('checked', true);
+                    downloadButton.dispatch('click');
+
+                    expect(windowSpy.mock.calls).toHaveLength(3);
+
+                    expect(windowSpy.mock.calls[0][0]).toContain('/iv/');
+                    expect(windowSpy.mock.calls[0][0]).toContain('sites=11112222');
```
mbucknell, line 102: I write the tests like this because the order of the parameters in the URL is meaningless and can change. Doing it this way also has the benefit of pinpointing what the issue.

assets/src/scripts/monitoring-location/store/hydrograph-data.js
```diff
@@ -97,7 +97,7 @@ const retrieveIVData = function(siteno, dataKind, {parameterCode, period, startT
         const isCalculatedTemperatureCode = isCalculatedTemperature(parameterCode);
 
         return fetchTimeSeries({
-            sites: [siteno],
+            siteno: siteno,
```
mbucknell, line 100: All of the fetch functions now take a single siteno and a single parameter code (which can be null). The code for this page only ever deals with a single site and we are either fetching a single parameter code or all of them. Having this mix of sometimes passing in a single siteno or sometimes an array of sitenos (and similarly for the parameter codes) is what caused the issue that I found when displaying groundwater levels when using custom date range (we were passing in a single siteno rather than an array). Also standardizing on siteno and parameterCode as the variable names

assets/src/scripts/web-services/groundwater-levels.js
```diff
@@ -7,18 +7,25 @@ import config from 'ui/config';
  *  Formats a URL for the purpose of downloading groundwater levels data in RDB format
  * @param  {String}  siteno - A code that uniquely identifies a monitoring location
  * @param  {String}  parameterCode - USGS five digit parameter code
- * @param {String} startTimeDT - ISO 8601 time
```
mbucknell, line 10: This is actually what caused the issue with the display of groundwater levels with custom time. Also standardizing the variable name as startTime/endTime as these are times (not just dates). We shouldn't let the misleading naming on the API leak into the code.

assets/src/scripts/web-services/groundwater-levels.js
```diff
@@ -7,18 +7,25 @@ import config from 'ui/config';
...
- * @param {String} endDT- ISO 8601 time
+ * @param {String} startTime - ISO 8601 time
+ * @param {String} endTime- ISO 8601 time
  * @param {String} format - The format of the data returned from the water services API, such as 'json' or 'rdb'
  * @return {String} The URL used to contact waterservices.usgs.gov
  */
-export const getServiceURLSGroundwater = function({siteno, parameterCode, period= null, startDT=null, endDT=null, format}) {
+export const getGroundwaterServiceURL = function ({
```
mbucknell, line 15: Also changed the name of the get*URL functions to follow the same pattern as above.

assets/src/scripts/web-services/groundwater-levels.test.js
```diff
@@ -20,30 +21,93 @@ describe('web-services/groundwater-levels', () => {
         restoreConsole();
     });
 
+    describe('getGroundwaterServiceURL', () => {
+        it('Expects site no and format parameters but no others', () => {
+            const result = getGroundwaterServiceURL({
+                siteno: '11112222',
+                format: 'rdb'
+            });
+            expect(result).toContain('sites=11112222');
```
mbucknell, line 30: This follows the same pattern as the tests in hydrograph/download-data.test.js. Allows order to change in the function implementation without breaking the tests.

assets/src/scripts/monitoring-location/store/hydrograph-data.js
abriggs-usgs: Wonderful!

assets/src/scripts/web-services/groundwater-levels.js
abriggs-usgs: Ah, good catch. I'm still trying to learn all the time/date formatting options.

assets/src/scripts/web-services/groundwater-levels.test.js
abriggs-usgs: Much less brittle.

assets/src/styles/components/hydrograph/_app.scss
```diff
@@ -55,6 +55,25 @@
...
+  @include grid-gap;
+  @include u-flex('row');
+  @include u-padding(2);
+  .usa-legend {
+    @include u-text('bold');
+  }
+  .download-info {
+    @include u-padding(2);
+    @include u-flex('auto');
+    @include u-flex('align-self-center');
```
abriggs-usgs, line 69: A very handy styling tool.

abriggs-usgs (approved):
Looks great! Very easy to understand and a good amount of cleaning and standardizing too.

assets/src/scripts/monitoring-location/components/hydrograph/download-data.js
```diff
@@ -0,0 +1,206 @@
...
+                            downloadUrl = getMedianDataURL(store, siteno);
+                            break;
+                        case 'groundwater-levels':
+                            downloadUrl = getGroundwaterLevelURL(store, siteno);
+                            break;
+                        case 'site':
+                            downloadUrl = getSiteMetaDataURL(siteno);
+                            break;
+                        default:
+                            console.log('Unexpected value');
```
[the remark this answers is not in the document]
mbucknell: Or even tomorrow! Very good point. I am changing this to a more meaningful message.

PR #1465, opened 2021-03-26
Wdfn 511 clarify na for wateralert

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-511 Clarify N/A for WaterAlert
-----------
Now that the WaterAlert subscription link is in the parameter list we have more room to add the title.
image

Also fixes a small bug in the link for the subscriptions for the converted temperature.
image

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
@@ -182,6 +182,14 @@ const drawRadioButtonRow = function(container, parameter, store) {
 * @param {D3 selection} parameter - Contains details about the current parameter code
 */
 const drawWaterAlertRow = function(container, siteno, parameter) {
+    // WaterAlert doesn't accept the calculated parameter for Fahrenheit (such as 00010F), so we need to adjust the
+    // parameter code back to the Celsius version (such as 00010).
+    const waterAlertURL = function() {
+        return parameter.parameterCode.includes(config.CALCULATED_TEMPERATURE_VARIABLE_CODE) ?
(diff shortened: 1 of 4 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -182,14 +182,22 @@ const drawRadioButtonRow = function(container, parameter, store) {
 * @param {D3 selection} parameter - Contains details about the current parameter code
 */
 const drawWaterAlertRow = function(container, siteno, parameter) {
+    // WaterAlert doesn't accept the calculated parameter for Fahrenheit (such as 00010F), so we need to adjust the
+    // parameter code back to the Celsius version (such as 00010).
+    const waterAlertURL = function() {
```
mbucknell, line 187: I'm wondering if we should include the waterAlertURL in the data returned from the selector? That would keep all of the Fahrenheit/Celsius code in a selector which is easier to test.

mbucknell (approved):

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: I like that idea, but how do we get the siteno in the selector?

abriggs-usgs: I went back and forth on this. I tried passing a function with the URL that accepted a site number, but that didn't seem great. Then, in looking more closely, I realized that the logic for the adjusting the calculated parameter code already was in the parameter-data and just wasn't making it to the right spot. So I compromised and used that to shorten the logic that creates the URL.

PR #1466, opened 2021-03-26
Wdfn 525 - Usability: Move custom time span selection

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Usability: Move custom time span selection

Description
-----------
This has been significantly redesigned from a UI perspective as shown below:

Screen Shot 2021-03-26 at 3 33 29 PM

I have decided we need to think more about whether we want to add a "years" text box to the Days before today section. Also I have not yet implemented the time span label. This should be easy to add when we decide where to put it.

The time span is now stored on a single property in the Redux store, hydrographState.selectedTimeSpan. If it is a string, it is assumed to be a duration of the form 'PXD' where X is a positive number. If a calendar date range, then the property value is an object with properties start and end.

After making a pull request
---------------------------


abriggs-usgs (approved):
Looks good to me. Clever use of 'typeOf' to differentiate between the input types!

PR #1474, opened 2021-03-29
Wdfn 520 estimated not rendering correctly

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-520 Estimated Not Rendering Correctly
-----------
The 'estimated' data points now have the expected color and legend lable.

Description
-----------
Changes here compensated for times with the data services return more than one qualifier code with the 'estimated' code.
The code now give priority to 'estimated', in those cases.

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/selectors/iv-data.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/selectors/iv-data.js
@@ -101,7 +101,11 @@ export const getIVDataPoints = memoize(dataKind => createSelector(
             let pointClass;
             const pointQualifiers = point.qualifiers.map(qualifier => qualifier.toLowerCase());
             const maskedQualifier = pointQualifiers.find(qualifier => qualifier in MASKED_QUALIFIERS);
-            const approvalQualifier = pointQualifiers.find(qualifier => qualifier in APPROVAL_QUALIFIERS);
+            // The data services sometimes return more than one approval code per point,
+            // such as ["A", "e"]. In these cases, give priority to 'e', which is 'estimated'.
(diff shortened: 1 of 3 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/selectors/iv-data.js
```diff
@@ -101,7 +101,11 @@ export const getIVDataPoints = memoize(dataKind => createSelector(
             let pointClass;
             const pointQualifiers = point.qualifiers.map(qualifier => qualifier.toLowerCase());
             const maskedQualifier = pointQualifiers.find(qualifier => qualifier in MASKED_QUALIFIERS);
-            const approvalQualifier = pointQualifiers.find(qualifier => qualifier in APPROVAL_QUALIFIERS);
+            // The data services sometimes return more than one approval code per point,
+            // such as ["A", "e"]. In these cases, give priority to 'e', which is 'estimated'.
+            const approvalQualifier = pointQualifiers.includes('e') ? 'e':
+                pointQualifiers.find(qualifier => qualifier in APPROVAL_QUALIFIERS);
```
mbucknell, line 107: So changing the order of APPROVAL_QUALIFIERS didn't work? I'm not opposed to this but think it could be brittle if more approvals are added (which isn't going to happen).

mbucknell (approved):

assets/src/scripts/monitoring-location/components/hydrograph/selectors/iv-data.js
abriggs-usgs: Unfortunately, changing the order didn't work. The qualifiers are in a regular object instead of an array with objects, so it only paid attention to the key names.

PR #1487, opened 2021-04-02
Endpoint name cleanup

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Endpoint name cleanup

Description
-----------
Cleaned up the names of the endpoints in the configuration. We are using separate config variables for each type of variable. Also removed some unused configuration parameters. This work is associated with WDFN-531 but not directly related to the issue in WDFN-531. Working on the service calls prompted me to clean up the naming.

After making a pull request
---------------------------


abriggs-usgs (approved):
Looks good to me. I like the standardization to 'endpoint' over 'root' or 'url'. The only name suggestion I would put out for discussion is 'PAST_IV_DATA_ENDPOINT'. The term 'past' is pretty vague. I might use something like 'ARCHIVED_IV_DATA_ENDPOINT', or some other term that indicates long term storage. But I'm okay with 'past' if you think that makes the most sense.

mbucknell:
Another option for  PAST_IV_DATA_ENDPOINT is HISTORICAL_IV_DATA_ENDPOINT. I have heard it referred to that way. Which do you prefer?

abriggs-usgs:
'Historical' sounds like a good option. I like it.

PR #1493, opened 2021-04-06
Wdfn 522 method picker moved

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-522 Method Picker Moved
-----------
Moves the sampling method selection into the parameter selection list.

Description
-----------
The option to select methods is now in the parameter selection list.

Things of note:
After trying out checkboxes, we found that their use, which would allow plotting multiple lines, would require some significant changes in code implementation and we thought it would be best to wait to see if users wanted this functionality before adding it. 

So, I changed to radio buttons. After implementing the radio buttons, it seemed the radio buttons had the same functionality as the dropdown list and took up significantly more space.

So, I reverted to the drop down list.

There are a few changes to the dropdown list.
1) The methods now appear in order from the one with the most points to the least. 
2) If a method has no points in the selected time span, it is disabled
3) If there is a method that is disabled, there is a message to indicating why it is disabled

cavates 
1) The methods are only available when a parameter is selected. So there is a slight delay before the method selection list shows on the screen.
2) Since the methods only are available with the selected parameter, no methods are available to show if the user expands a row but doesn't select it. 
3) The dropdown list isn't perfect on mobile devices, since the options will not 'row wrap'. Long labels are truncated. 

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
```diff
@@ -28,37 +34,57 @@ const updateAvailableMethods = function(selectElem, methods, store) {
...
-        selectElem.append('option')
-            .text(method.methodDescription ? `${method.methodDescription}` : 'None')
+        const hasPointsInTimeSpan = method.pointCount > 0;
+        container.append('option')
+            .text(`${method.methodDescription}`)
+            .attr('class', 'method-option sampling-method-selection')
             .attr('selected', method.methodID === selectedMethodID ? true : null)
-            .node().value = method.methodID;
+            .attr('value', method.methodID)
+            .property('disabled', !hasPointsInTimeSpan ? true : null);
```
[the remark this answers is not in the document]
mbucknell: In other places we use the attribute not the property for 'disabled'. Is there are reason here? If it is a property than you should pass in boolean (then it should be .property('disabled', !hasPointsInTimeSpan).

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
```diff
@@ -28,37 +34,57 @@ const updateAvailableMethods = function(selectElem, methods, store) {
...
+ * @param {Redux store} store - A complex JavaScript object that maps the current state of the application.
  */
-export const drawMethodPicker = function(elem, store) {
-    const pickerContainer = elem.append('div')
+export const drawMethodPicker = function(container, parameterCode, store) {
+    const gridRowSamplingMethodSelection = container.append('div')
+        .attr('id', 'primary-sampling-method-row')
+        .attr('class', 'grid-container grid-row-inner sampling-method-selection');
+    const methodSelectionContainer = gridRowSamplingMethodSelection.append('div')
+        .attr('id', `method-selection-container-${parameterCode}`)
```
mbucknell, line 60: Is there a reason we need this id? I think we should try not to put ids or classes on things unless needed. It just increases the cognitive load. In the above example for instance, you have the class "sampling-method-selection". That class is used for divs (at different levels), options, and labels.  I see that you are using it in parameters.js but I don't think it needs to be on all of these things (probably just the "option" tag)

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
```diff
@@ -28,37 +34,57 @@ const updateAvailableMethods = function(selectElem, methods, store) {
...
+ * @param {D3 selection} container - An HTML element to serve as the base for attaching other elements.
+ * @param {String} parameterCode - The five digit USGS that is unique for each parameter.
+ * @param {Redux store} store - A complex JavaScript object that maps the current state of the application.
  */
-export const drawMethodPicker = function(elem, store) {
-    const pickerContainer = elem.append('div')
+export const drawMethodPicker = function(container, parameterCode, store) {
+    const gridRowSamplingMethodSelection = container.append('div')
+        .attr('id', 'primary-sampling-method-row')
+        .attr('class', 'grid-container grid-row-inner sampling-method-selection');
```
mbucknell, line 58: Can we use mixins rather than utility classes for grid-container, grid-row (below). Is grid-row-inner needed (we can't do the styling with standard USWDS classes/mixins)?

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -154,7 +187,7 @@ const drawRadioButtonRow = function(container, parameter, store) {
     const gridRowInnerWithRadioButton = container.append('div')
         .attr('class', 'grid-row grid-row-inner');
```
mbucknell, line 188: Can we try to use mixins rather than utility classes for grid-row?

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
```diff
@@ -248,10 +281,17 @@ export const drawSelectionList = function(container, store, siteno) {
             .attr('id', `expansion-container-row-${parameter.parameterCode}`)
             .attr('class', 'expansion-container-row')
             .attr('hidden', 'true');
+
         // Add the rows nested in the expansion container
         if (parameter.waterAlert.hasWaterAlert) {
             drawWaterAlertRow(expansionContainerRow, siteno, parameter);
         }
+
+        // Add the sampling method selection list to the parameter loaded when the page is first opened
```
mbucknell, line 290: nit: Do you mean "row is first opened"?

assets/src/scripts/monitoring-location/components/hydrograph/selectors/time-series-data.js
```diff
@@ -150,3 +150,34 @@ export const getPreferredIVMethodID = createSelector(
...
+        if (!ivData || !Object.keys(ivData.values).length) {
+            return null;
+        }
+        const methodsForPrimarySelection = Object.values(ivData.values)
+            .map(methodValues => {
+                return {
+                    pointCount: methodValues.points.length,
+                    lastPoint: methodValues.points.length ? methodValues.points[methodValues.points.length - 1] : null,
+                    methodID: methodValues.method.methodID,
+                    methodDescription: methodValues.method.methodDescription
```
mbucknell, line 171: Not all methods have a description. We need to put some sort of string here. We were using "None"

assets/src/scripts/monitoring-location/components/hydrograph/selectors/time-series-data.js
```diff
@@ -150,3 +150,34 @@ export const getPreferredIVMethodID = createSelector(
...
+                };
+            })
+            .sort((a, b) => {
+                if (a.pointCount === b.pointCount) {
+                    return a.pointCount ? a.lastPoint.dateTime - b.lastPoint.dateTime : 0;
+                } else {
+                    return a.pointCount - b.pointCount;
+                }
+            });
+        return methodsForPrimarySelection.reverse();
```
mbucknell, line 181: You should be able to write sort to return the reverse order (so "reverse" is not needed). Does this work:
```
if (a.pointCount === b.pointCount) {
    return a.pointCount ? b.lastPoint.dateTime - a.lastPoint.dateTIme : 0;
} else {
    return b.pointCount - a.pointCount
}

mbucknell (approved):
I do have a bunch of comments and suggestions and make sure the old markup for the method picker (and any styling) is removed.

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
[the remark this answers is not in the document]
abriggs-usgs: Very good point. I always confuse the hidden property and the disabled attribute and which is the one is property vs attribute. The boolean works more easily with the attribute too.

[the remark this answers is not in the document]
mbucknell: They are both attributes. I believe selected and value (at least for text input's) are properties because those can be set by the browser.

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
abriggs-usgs: Good eye. Yes, some of this is left over because I was having a devil of a time getting the '.remove()' to work. I'll see if I can clean this up.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: Well, kinda. I added a bit more explanation, not sure if it helps. This code deals with the fringe case when the page loads and the default parameter has methods, but the user never actually clicks the row, but uses the row expansion toggle (simple no? :) ) .       

This is what I added:
 // Add the sampling method selection list to the parameter loaded when the page is first opened if it has methods.
 // This allows the user to use the row expansion toggle to see the methods when the row has never been physically clicked.

assets/src/scripts/monitoring-location/components/hydrograph/selectors/time-series-data.js
abriggs-usgs: There is code that catches that case further downstream, but you are right, and it would make more sense for that to happen here.

assets/src/scripts/monitoring-location/components/hydrograph/selectors/time-series-data.js
abriggs-usgs: Ah, neat. That saves a little computing.

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
mbucknell: You might want to put a clarifying comment on the spot where you do the exclude.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
mbucknell: Okay. There does seem to be some "magic" around all of the clicking/expanded/contracting. I'm afraid that this code will be somewhat brittle. Hopefully I am wrong.

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.js
abriggs-usgs: Can do!

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: True. If we had all of the sampling method options for every parameter when the table was created things would be less complicated. So that might be a motivating factor to making a change to the data reducer as some future point.

assets/src/scripts/monitoring-location/components/hydrograph/parameters.js
abriggs-usgs: Darn, forgot to change those. I have to be honest the mixins are neat, but I find it hard to set up the initial structure of grids and similarly complex CSS  using them from the start because it makes it harder for me to see what grid row is holding which columns. So I tend to set things up with the utility classes and than convert them over. And I always seem to miss a few.

abriggs-usgs: Well, I thought about this over the night, and I'm not sure if using the @include for the 'grid-rows' would make things more clear. Since the 'grid-row' is the only class on these divs (yes you were right, the 'grid-row-inner is not used), I don't have a good way to identify them in the CSS so that the @include can be used. If I kept the 'grid-row-inner', I could use that as the selector and attach the @include to that, but that seems like I'm only trading one class for another.

PR #1496, opened 2021-04-09
Wdfn 531 - Use requests Session in waterdataui server

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Use requests Session in waterdataui server

Description
-----------
The four services that are used to retrieve data for the monitoring-location and networks pages now use requests.Session which should reused connections and hopefully boost performance. I choose not to update the camera service and that is only called once when the applications starts but we may want to consider that in the future.

After making a pull request
---------------------------


wdfn-server/waterdata/tests/test_views.py
```diff
@@ -221,35 +210,32 @@ def setUp(self):
...
-    @mock.patch('waterdata.services.ogc.get_networks')
+    @mock.patch('waterdata.services.ogc.MonitoringLocationNetworkService.get_networks')
     def test_invalid_network(self, network_mock):
-        network_mock.return_value.status_code = 200
-        network = 'monitoring-locations-invalid'
-        url = '/networks/{}/'.format(network)
-        response = self.app_client.get(url)
+        network_mock.return_value = {}
+        response = self.app_client.get('/networks/')
         # Assert an invalid network in gui does not 404
```
abriggs-usgs, line 231: Is this comment still correct? Seems like the opposite is happening.

mbucknell: Thanks. I meant to delete this test.

PR #1505, opened 2021-04-12
Wdfn 532 clean up page when site has no ivgw data

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-532 Clean Up Page When Site has No Instantaneous Value (IV) or Ground Water (GW) Data
-----------
Removes empty elements on pages with no IV or GW data

Description
-----------
Removes the provisional data statement and change time span button.  Also allows the 'retrieve data' button to show (previously prevented by null object error) to show on on pages with no IV or GW data.

After making a pull request
---------------------------


assets/src/scripts/monitoring-location/components/hydrograph/index.test.js
```diff
@@ -318,10 +318,6 @@ describe('monitoring-location/components/hydrograph module', () => {
             expect(selectAll('#change-time-span-container').size()).toBe(1);
         });
 
-        it('should have download data form', () => {
```
mbucknell, line 321: Why was this test removed? We still have a download-graph-data-container?

assets/src/scripts/monitoring-location/components/hydrograph/select-actions.test.js
```diff
@@ -38,8 +42,8 @@ describe('monitoring-location/components/hydrograph/select-actions', () => {
         expect(buttonContainer.size()).toBe(1);
         expect(buttonContainer.selectAll('button').size()).toBe(2);
 
-        const downloadButton = buttonContainer.selectAll(`button[aria-controls=${TIME_FORM_ID}]`);
-        const timeSpanButton = buttonContainer.selectAll(`button[aria-controls=${DOWNLOAD_FORM_ID}]`);
```
mbucknell, line 42: Oops!

assets/src/scripts/monitoring-location/components/hydrograph/select-actions.test.js
```diff
@@ -106,4 +110,51 @@ describe('monitoring-location/components/hydrograph/select-actions', () => {
         expect(testDiv.select(`#${TIME_FORM_ID}`).attr('hidden')).toBeNull();
         expect(testDiv.select(`#${DOWNLOAD_FORM_ID}`).attr('hidden')).toBe('true');
     });
-});
\ No newline at end of file
+
+    describe('monitoring-location/components/hydrograph/select-actions with no iv or gw data', () => {
```
mbucknell, line 114: This isn't exactly what I meant by nesting describe sections. You are actually creating the select actions element twice for the nested describe because the beforeEach for the parent describe will also get executed for each test. I think it would be clearer to have:
```
describe('monitoring-location/components/hydrograph/select-actions', () => {
    #Define constants that are common but not TEST_STATE which will be different for the two sub-suites
    describe('Tests when iv or gw data is available', () => {
         #Add everything in the current describe except not the the nested describe
    describe('Tests with no IV or GW data', () => {
        # Add everything in the current nested describe.

mbucknell (approved):
See comments on the tests. I think they need to be changed or we may have instances where the tests sometimes fail.

assets/src/scripts/monitoring-location/components/hydrograph/index.test.js
abriggs-usgs: Whoops.  My bad. At one point I modified this test. Then when removing the modification, I apparently got a little carried away with my deletions. Good catch!

assets/src/scripts/monitoring-location/components/hydrograph/select-actions.test.js
abriggs-usgs: Ah, that makes so much more sense!! Thanks for pointing that out.

PR #1531, opened 2021-05-05
show method id

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WDFN-546 Methods with No Description Should Show Method ID
-----------
The sampling method names are sketchy at best and sometimes not even there. This defaults the methods descriptions that are not there to their method ID.

It also fixes a small bug where when the time span was changed, the method note was rendered under the method selection element rather than above. 

After making a pull request
---------------------------

### DIFF
--- a/assets/src/scripts/monitoring-location/components/hydrograph/method-picker.test.js
+++ b/assets/src/scripts/monitoring-location/components/hydrograph/method-picker.test.js
@@ -141,5 +141,34 @@ describe('monitoring-location/components/hydrograph/method-picker', () => {
 
             expect(div.select('#no-data-points-note').size()).toBe(0);
         });
+
+        it('Expects that if there is not method description, the method id will be used', () => {
+            const parameterCode = '72019';
+            let store = configureStore({
+                ...TEST_STATE,
+                hydrographData: {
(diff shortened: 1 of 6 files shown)


assets/src/scripts/monitoring-location/components/hydrograph/method-picker.test.js
```diff
@@ -141,5 +141,34 @@ describe('monitoring-location/components/hydrograph/method-picker', () => {
 
             expect(div.select('#no-data-points-note').size()).toBe(0);
         });
+
+        it('Expects that if there is not method description, the method id will be used', () => {
```
mbucknell, line 145: I think the more appropriate place for this test is in time-series-data. After all that is where the logic is. So I recommend this test essentially be moved to time-series-data.test.js

mbucknell (approved):
See comment for a suggestion on testing.

assets/src/scripts/monitoring-location/components/hydrograph/method-picker.test.js
abriggs-usgs: Sounds reasonable. Done!

